PR #10, opened 2017-08-19
Added Decimal to BInary Conversion

ms10398:


r0hit-gupta:
Great Work!

ms10398:
Looking forward to add more to the organisation,
Will be happy to be a member of it so that no need to make PR's!!

Issue #14, opened 2017-09-28
Add more algorithms


AnupKumarPanwar:
Please add more algorithms to this repo. 

See the structure and README of this repo for reference
https://github.com/TheAlgorithms/Python

meetzaveri:
#18

kulshreshtha19:
[bubblesort.js.txt](https://github.com/TheAlgorithms/Javascript/files/1821995/bubblesort.js.txt)
Uploading bubblesort algorithm in javascript. kindly merge it.
Thank you

ntedgi:
#84

raviolliii:
#83 Added SHA-1 and SHA-256 Hashing algorithms

Satzyakiz:
I would like to work on this issue.

itsvinayak:
#195

TangibleDream:
I added GridGetX and GridGetY #222

harshitagrawal7733:
What to do? 
Add more alogorithms is the goal?

itsvinayak:
@harshitagrawal7733 adding algorithms which follow coding standards of repo 
please have a look [here](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md)

dev348:
Yeah sure 👍🏾

aakhtar3:
I just added a few algorithms.

Are we able to add diagrams? I have a bunch of notes that tie to my code to help understand the data structures and algorithms.

vinay72:
@AnupKumarPanwar Hi, Anup. I've added a De-facto unbiased shuffle algorithm also known as the Fisher-Yates shuffle algorithm, so please review and merge my PR.

arjun1311:
Hi, I have added an algorithm please review and merge my PR

allenjamesvinoy14:
Can you please open an issue with hacktoberfest label to add data structures as well? Or may I add a data structure under this issue?

sonechca:
Hi, I added a new algorithm "Tree Sort".
Can you check my code?
Thank you https://github.com/TheAlgorithms/Javascript/pull/513

elouie-code:
Added a linked list reverse order algorithm. can you review please? [#528](https://github.com/TheAlgorithms/Javascript/pull/528)

Thanks!

achudinovskyh:
Hi, I've added new Search algorithm, can you check it please? Thank you! #542

PR #18, opened 2017-09-29
Binary search tree

meetzaveri:
I have added binary search tree algorithm for javascript
#14


ms10398:
There is already a BInary Search Tree in data structures
You should modify it instead of adding a new file.
Thank you and if modifying then list the changes so that it is easy to review

PR #36, opened 2017-10-19
Create quickSort.js

LukeSchlangen:


ms10398:
It is already there @LukeSchlangen you should check the sorts section.

PR #64, opened 2018-10-16
Implemented bucket sort algorithm

KuLi:
Added bucket sort algorithm


Sorts/bucketSort.js
```diff
@@ -0,0 +1,62 @@
...
+*/
+function bucketSort(list, size){
+    
+    if(undefined === size){
+        size = 5;
+    }
+    if(list.length === 0){
+        return list;
+    }
+    var min = list[0];
```
ms10398: Can you please use ES6 instead of `var`

```suggestion
    let min = list[0];
```

KuLi:
Sure. I changed it.

PR #67, opened 2018-10-20
feat: add Strings section with checkPalindrome algorithm

v1ktoryy:
Issue related: #14


ms10398:
As stated in the previous PR we are focusing more on algorithms than language features

PR #68, opened 2018-10-20
feat: StingsIntegerReversal

v1ktoryy:
Issue related to: #14


ms10398:
@ParthS007 What do you say I am not sure if a algorithm should allow just everything to be done by Library functions while working on not any algorithm.

It looks more like a trick solution :wink: I need your views and should we go with merging PRs like this in future.

PR #83, opened 2019-05-01
Added SHA-1 and SHA-256 hashing algorithms

raviolliii:
Implemented the SHA-1 and SHA-256 hashing algorithms in Javascript.

### DIFF
--- a/Hashes/SHA1.js
+++ b/Hashes/SHA1.js
@@ -45,7 +45,7 @@ function chunkify(str, size) {
 }
 
 /**
- * Rotates string represention of bits to th left
+ * Rotates string representation of bits to the left
  *
  * @param {string} bits - string representation of bits
  * @param {int} turns - number of rotations to make
@@ -65,7 +65,7 @@ function rotateLeft(bits, turns) {
  * @return {string} - processed message
  */
 function preProcess(message) {
-    //covert message to binary representation padded to
+    //convert message to binary representation padded to
     //8 bits, and add 1
     let m = message.split("")
         .map(e => e.charCodeAt(0))
--- a/Hashes/SHA1.js
+++ b/Hashes/SHA1.js
@@ -1,0 +1,183 @@
+//================================================================
+// SHA1.js
+//
+// Module that replicates the SHA-1 Cryptographic Hash 
+// function in Javascript.
+//================================================================
+
+//main variables
+const CHAR_SIZE = 8;
+
+/**
+ * Adds padding to binary/hex string represention
+ *
+ * @param {string} str - string represention (binary/hex)
+ * @param {int} bits - total number of bits wanted
+ * @return {string} - string represention padding with empty (0) bits
+ *
+ * @example
+ *      pad("10011", 8); // "00010011"
+ */
+function pad(str, bits) {
+    let res = str;
+    while (res.length % bits !== 0) {
+        res = "0" + res;
+    }
+    return res;
+}
+
+/**
+ * Separates string into chunks of the same size
+ *
+ * @param {string} str - string to separate into chunks
+ * @param {int} size - number of characters wanted in each chunk
+ * @return {array} - array of original string split into chunks
+ *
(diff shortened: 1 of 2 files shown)


Hashes/SHA1.js
```diff
@@ -0,0 +1,183 @@
...
+function chunkify(str, size) {
+    let chunks = [];
+    for (let i = 0; i < str.length; i += size) {
+        chunks.push(str.slice(i, i + size));
+    }
+    return chunks;
+}
+
+/**
+ * Rotates string represention of bits to th left
```
siriak:
```suggestion
 * Rotates string representation of bits to the left
```

Hashes/SHA1.js
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * Pre-processes message to feed the algorithm loop
+ *
+ * @param {string} message - message to pre-process
+ * @return {string} - processed message
+ */
+function preProcess(message) {
+    //covert message to binary representation padded to
```
siriak:
```suggestion
    //convert message to binary representation padded to
```

Hashes/SHA1.js
```diff
@@ -0,0 +1,183 @@
...
+    
+    chunks.forEach(function(chunk, i) {
+        //break each chunk into 16 32-bit words
+        let words = chunkify(chunk, 32);
+
+        //extend 16 32-bit words to 80 32-bit words
+        for (let i = 16; i < 80; i++) {
+            let val = [words[i - 3], words[i - 8], words[i - 14], words[i - 16]]
+                .map(e => parseInt(e, 2))
+                .map(e => e >>> 0)
```
siriak: Could you please explain why do we need `.map(e => e >>> 0)` ?  Doesn't e already fits into 32 bit?

Hashes/SHA1.js
```diff
@@ -0,0 +1,183 @@
...
+                .reduce((acc, curr) => curr ^ acc, 0);
+            let bin = (val >>> 0).toString(2);
```
siriak: The same question here. As I understand, numbers here are already 32 bit unsigned and ^ operator cannot extend them.

Hashes/SHA1.js
raviolliii: Since the words in each chunk are stored as bitstring representations, they must be converted to integers using `parseInt`. As I understand it, integers in Javascript are signed by default, so before using the xor operator (`^`), they must be converted to unsigned ints using `>>> 0`.

Hashes/SHA1.js
raviolliii: Same reasoning here as mentioned above, after using the xor operator on the words, the resulting value (`val`) can still be interpreted by Javascript as an signed int (as I understand it/tested it), so it's converted to unsigned, then the string version.

Hashes/SHA1.js
siriak: The reason why `e >>> 0` is not required here is the way `parseInt` function works. It doesn't view leading 1 or 0 as sign bit, instead it decides that number is negative if the string starts from minus '-' sign. Thus in our case `e` is always positive.

Hashes/SHA1.js
siriak: I see your point here.
`parseInt("01111111111111111111111111111111", 2) ^ parseInt("11111111111111111111111111111111", 2)`
gives negative number.

Hashes/SHA1.js
raviolliii: Oh okay I see now, thank you for the explanation. I'll remove the line and commit, as well as resolve the comment below.

raviolliii: I'm sorry, this is my first pull request. I'm just wondering if I would now merge this PR or whether someone else with merge permissions would. Sorry for the confusion.

siriak: Sorry for the delay, I will merge it right now :)

PR #90, opened 2019-08-20
Add operator *= for function factorial. This is easy to understand.

gerssonmg:
Add operator *= for function factorial. This is easy to understand.


AkarshSatija:
When It comes to understanding, `factorial = factorial * i;` is better to understand

PR #116, opened 2019-12-11
docs: added navigation for better readability

shrmaky:
- [x]  Added Navigation for all the Algorithms mentioned in README.md
- [x] Passing all checks 

Two issues -
- Line#113 contains a link Complexity Graphs which is not working
- Line#116 Have to change the link name from 'Vigenére'(read é) to 'Vigenere' because the hash doesn't links, any help on this is appreciated.


itsvinayak:
we have changed README.md Please have a look, and add you change to the wiki,
thank you

PR #125, opened 2020-04-17
Graph Algorithm

Nur69:


Nur69:
okay, i will see how to fix it 
thank you.

PR #128, opened 2020-04-26
Add doublylinkedlist

hmizz:
i added a doubly-linked-list function to the linked list directory!


hmizz:
i removed the unused variables

hmizz:
the path in the directory was changed as requested

PR #138, opened 2020-05-03
Apply the JavaScript Standard Style

cclauss:
Drive contributions to use a common code style...  https://standardjs.com

This PR was created with the commands:
```
npm install standard --save-dev  # Required just once
npx standard --fix
```

Why not pick a different code formatter?  Like `gofmt` and `psf/black`, the `standard` formatter is ___not configurable___ (this is a feature, not a bug) in an attempt to create a common code format across all projects written in their given language.


ParthS007:
@cclauss We can go with `standard`.

Issue #139, opened 2020-05-03
Files that are not compliant with standardjs


cclauss:
The following FilePaths fail `npx standard FilePath` and can not be fixed using the `--fix` option.
* https://standardjs.com
* [x] Javascript/linear-algebra-javascript/test/test.js
* [x] Javascript/linear-algebra-javascript/src/la_lib.js
* [ ] Remove [`|| true` from the `npx standard` command](https://github.com/TheAlgorithms/Javascript/blob/master/.github/workflows/nodejs.yml#L23)

---
* [x] Javascript/Ciphers/keyFinder.js #140
* [x] Javascript/Ciphers/vigenereCipher.js  #140
* [x] Javascript/Conversions/DecimalToHex.js  #140
* [x] Javascript/Data Structures/Graph/Graph.js #141
* [x] Javascript/Data Structures/Heap/MinPriorityQueue.js  #140
* [x] Javascript/Data Structures/Linked List/DoublyLinkedList.js  #140
* [x] Javascript/Data Structures/Queue/Queue.js  #140
* [x] Javascript/Data Structures/Stack/Stack.js  #140
* [x] Javascript/Data Structures/Tree/Binary Search Tree.js  #140
* [x] Javascript/Hashes/SHA256.js
* [x] Javascript/Search/linearSearch.js #143
* [x] Javascript/Sorts/TopologicalSort.js #141
* [x] Javascript/Sorts/bogoSort.js #150
* [x] Javascript/Sorts/bucketSort.js #141
* [x] Javascript/Sorts/combSort.js #141
* [x] Javascript/Sorts/cycleSort.js #141
* [x] Javascript/Sorts/heapSort.js  #150
* [x] Javascript/Sorts/selectionSort.js #141
* [x] Javascript/Sorts/wiggleSort.js  #150
* [x] Javascript/maths/DijkstraSmallestPath.js #147
* [x] Javascript/maths/abs.js #144 
* [x] Javascript/maths/factorial.js #145 
* [x] Javascript/maths/find_lcm.js #146
* [x] Javascript/maths/graph.js

juliano-nunes:
Can I help with `Javascript/Ciphers/vigenereCipher.js` ?

juliano-nunes:
Hey @itsvinayak, let me know if you need help :)

cclauss:
[...]

PR #140, opened 2020-05-04
Fixing non compliant files (Ciphers, Conversions and Data Structures)

juliano-nunes:
Resolve some of the topics in #139


itsvinayak:
please add a calling function in Graph.js, so that Data Structures is fixed

PR #150, opened 2020-05-05
sort/

itsvinayak:
/* eslint no-extend-native: ["off", { "exceptions": ["Object"] }] */

this can be used to turn off eslint, as according to eslint extending native objects is bad

### DIFF
--- a/Sorts/bogoSort.js
+++ b/Sorts/bogoSort.js
@@ -2,6 +2,8 @@
  * A simple helper function that checks, if the array is
  * sorted in ascending order.
  */
+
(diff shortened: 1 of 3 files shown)


cclauss:
Not a fan of this approach.  The linter is correct that Array prototype is read only.

Can't we find a way to resolve this that does not cause the linter to complain.

itsvinayak:
Yes, but we have to change the entire code
We have to pass array and than sort
I guess




Vinayak sharma

9259888898

IERT allahabad
CSE 3rd year

Issue #164, opened 2020-05-10
Rationalize the two different implementations of heap sort


cclauss:
We have [two different implementations of heap sort](https://github.com/TheAlgorithms/Javascript/tree/master/Sorts).  Can we build a benchmark to prove which one is faster or better than the other?  Benchmark against small datasets, large datasets, and empty datasets.  It is OK to keep both if they have different strengths and weaknesses.  If there is a clear winner, let's delete the other one.

sahilbansal17:
@cclauss It seems that both the implementations are exactly the same algorithmically.

cclauss:
> Can we build a benchmark to prove which one is faster or better than the other?

It would be cool to have a benchmark because then we would have ___proof___ that they are the same and have the same performance for multiple datasets.  Hint: I doubt it.

Such a benchmark could be extended to do side-by-side performance tests of all 19 of our sort algorithms.  Benchmarks are cool and they have the power to push our contributors to optimize.

poyea:
Also, there's something more we can do. At least [BogoSort](https://github.com/TheAlgorithms/Javascript/blob/master/Sorts/BogoSort.js#L7) has an `isSorted` function and a `shuffle` function. We can make use of these two functions to build our test sets for the whole `sort/`.

cclauss:
;-)  Are lines 10-13 needed?

poyea:
> ;-) Are lines 10-13 needed?

If the loop starts with `i=1` and compares `this[i-1]` with `this[i]` then it is, but no for the current code.

cclauss:
Does the loop ever execute at all?

poyea:
Yes. It's kind of an extension for the Array object. That's why we can use it directly.
https://js.do/code/438412

cclauss:
Our test cases are when `length < 2` so...
document.write([].isSorted());  // --> true
document.write([1].isSorted());  // --> true

Now comment out lines 10-13 and run again.

You will get the exact same results because the loop never runs.

Lines 10-13 can be deleted.

PR #167, opened 2020-05-10
run: node .github/workflows/UpdateDirectory.js 

### LINKED ISSUE #160: Build DIRECTORY.md with JavaScript, not Python
We present all of of our algorithms in an organized list, [__DIRECTORY.md__](../blob/master/DIRECTORY.md) so visitors can rapidly find topics of interest.  Currently DIRECTORY.md is being autogenerated by one of [our GitHub Actions](../tree/master/.github/workflows) which [uses Python code](../blob/master/.github/workflows/update_directory_md.yml#L11-L57) to create the directory text.

Why should the JavaScript repo rely on Python code?!?  Especially given that GitHub Actions fully support the [creation of JavaScript-based Actions](https://help.github.com/en/actions/building-actions/creating-a-javascript-action).  Can we use JavaScript to autogenerate DIRECTORY.md and lose the Python code in the process?

cclauss:

Fixes #160

### DIFF
--- a/.github/workflows/update_directory_md.yml
+++ b/.github/workflows/update_directory_md.yml
@@ -7,9 +7,8 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - uses: actions/setup-node@v1
-      - run: node UpdateDirectory.js
-      - name: Update DIRECTORY.md
-        run: |
+      - run: |
+          node .github/workflows/UpdateDirectory.js
           cat DIRECTORY.md
           git config --global user.name github-actions
           git config --global user.email '${GITHUB_ACTOR}@users.noreply.github.com'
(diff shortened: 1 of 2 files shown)


cclauss:
@sahilbansal17 Your review please...  Your script seems to be doing its magic just fine!

cclauss:
@poyea Thanks for your review.

The heap sort thing is covered in #164 where I am advocating that we have a __benchmark__ that determines which implementation lives and which one dies.

It is bugging me that we have a directory called __algorithm__ because every submission to this repo should be an algorithm.  We should move the files in that directory to other dirs that provide more specificity to what _kind_ of algorithm it is.

QuickSelect.js  --> Search make sense to me.

PR #209, opened 2020-06-27
Update BinarySearch.js

Askanders:
The old algorithm didn't work, I believe for two main reasons:
1 -  Number.MAX_VALUE is not a valid array index as it is used to represent the highest possible value in javascript (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE);

2 - splice() is not a pure function, every time it is called it has the side effect of modifying the original array (https://www.w3schools.com/jsref/jsref_splice.asp) ;

So I rewrote the algorithm, it now returns an index ( -1 if not found ) and it works both on numbers and on strings.

### **Describe your change:**

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [x] Documentation change?

### **Checklist:**

### DIFF
--- a/Search/BinarySearch.js
+++ b/Search/BinarySearch.js
@@ -7,7 +7,7 @@
  * value is found or the interval is empty.
  */
 
-function binarySearch (arr, x, low = 0, high = arr.length - 1){
+function binarySearch (arr, x, low = 0, high = arr.length - 1) {
   const mid = Math.floor(low + (high - low) / 2)
 
   if (high >= low) {
--- a/Search/BinarySearch.js
+++ b/Search/BinarySearch.js
@@ -7,25 +7,25 @@
  * value is found or the interval is empty.
  */
 
-function binarySearch(arr, x, low = 0, high = arr.length - 1) {
-  let mid = Math.floor(low + (high - low) / 2);
+function binarySearch( arr, x, low = 0, high = arr.length - 1 ) {
+  const mid = Math.floor(low + (high - low) / 2)
 
   if (high >= low) {
     if (arr[mid] === x) {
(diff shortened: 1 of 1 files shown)


Search/BinarySearch.js
```diff
@@ -1,26 +1,70 @@
...
-  var mid = Math.floor(arr.length / 2)
-  if (arr[mid] === i) {
-    console.log('match', arr[mid], i)
-    return arr[mid]
-  } else if (arr[mid] < i && arr.length > 1) {
-    binarySearch(arr.splice(mid, Number.MAX_VALUE), i)
-  } else if (arr[mid] > i && arr.length > 1) {
-    binarySearch(arr.splice(0, mid), i)
+function binarySearch(arr, x, low = 0, high = arr.length - 1) {
+  let mid = Math.floor(low + (high - low) / 2);
```
ruppysuppy:
```suggestion
  const mid = Math.floor(low + (high - low) / 2);
```

ruppysuppy:
@Askanders, you can run ```$ standard MyFile.js``` as stated in [CONTRIBUTING.md](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md) to format the file instead of formatting it manually

Askanders:
Yes, thank you @ruppysuppy. I have a lot of configuration to do so I haven't installed the package yet, but I'll definitely do it in the future.

itsvinayak:
> Everything else looks fine to me, still I would appriciate if @itsvinayak would review it once

sure and thanks for helping @ruppysuppy

PR #237, opened 2020-08-01
Adding TimingFunctions/IntervalTimer

nandanvasudevan:

### **Describe your change:**

* [x] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [ ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -113,6 +113,9 @@
   * [ReverseString](https://github.com/TheAlgorithms/Javascript/blob/master/String/ReverseString.js)
   * [ReverseWords](https://github.com/TheAlgorithms/Javascript/blob/master/String/ReverseWords.js)
 
+## Timing Functions
+  * [IntervalTimer](https://github.com/TheAlgorithms/Javascript/blob/master/TimingFunctions/IntervalTimer.js)
+
 ## Trees
   * [DepthFirstSearch](https://github.com/TheAlgorithms/Javascript/blob/master/Trees/DepthFirstSearch.js)
 
--- a/TimingFunctions/IntervalTimer.js
+++ b/TimingFunctions/IntervalTimer.js
@@ -37,7 +37,7 @@ class IntervalTimer {
    */
   resetTimer () {
     clearInterval(this.timer)
-    this.callBack = () => {};
+    this.callBack = () => {}
     return this.getElapsedTime()
   }
 
--- a/TimingFunctions/IntervalTimer.js
+++ b/TimingFunctions/IntervalTimer.js
@@ -1,0 +1,93 @@
+/**
+ * @author Nandan V
+ * Sunday, 26 July 2020 8:33 AM
+ * @description Singleton class that handles the <b>timing of tests</b> and
+ *   specs. <br/> The class is singleton as <u>javascript does not support
+ *   multiple timer instances<u/>.
+ */
+class IntervalTimer {
+  /**
+   * @description Constructor for Timer.
+   * @param interval Sets the interval for running the timer.
+   * @param callBack The callback function to be executed.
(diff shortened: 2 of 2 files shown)


itsvinayak:
@nandanvasudevan thanks for contributing,

Please read Contribution Guidelines and run `standard --fix` command
Thank you
@itsvinayak

Sheraff:
This is all kinds of wrong. 

Your class isn't a singleton because `this.instance` always refers to `this` and not the class itself.

``` javascript
a = new IntervalTimer()
b = new IntervalTimer()
console.log(a === b) // false
```

A way to make a singleton that is close to what you wrote would be 

``` javascript
class IntervalTimer {
  constructor() {
    if(IntervalTimer.instance)
      return IntervalTimer.instance
    // ... rest of constructor here
  }
}
```

Also, there is no reason to make it a singleton since JS supports very well having multiple `setInterval` running concurrently, so your description is wrong:

> The class is singleton as javascript does not support multiple timer instances

You don't actually measure time elapsed: `setInterval` returns an ID meant to be passed to `clearInterval` and isn't specifically related to any time measurements. This logic is wrong in `getRunTime` as well as in `getElapsedTime`:

``` javascript
startTimer () {
  this.timer = setInterval(this.callBack, this.interval)
}
/**
  * @return {number} Elapsed time since start.
  */
getRunTime () {
  return this.timer
}
``` 

On a more conceptual level, `setInterval` isn't particularly well suited to keep track of time because it can be delayed arbitrarily. One should prefer the [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance) if possible (ES, but not Node).

The best pure-JS interval will always have to be blocking (in which case making it a singleton makes sense), but that's only useable in a threaded context (like a web worker or worker thread) because it blocks the execution of anything else.

``` javascript
class BlockingInterval {
    constructor(
        callback = () => {}, 
        interval = 0
    ) {
        if(BlockingInterval.instance)
            return BlockingInterval.instance

        this.callback = callback
        this.interval = interval
        this.click(this.interval)
    }

    click(interval, start = performance.now()) {
        while(performance.now() - start < interval){}
        this.callback()
        this.click(interval, start + interval)
    }
}

let last = performance.now()
const a = new BlockingInterval(() => {
    const current = performance.now()
    console.log(current - last)
    last = current
}, 500)
```

If you need an interval for drawing in the browser, use the same principle but with `requestAnimationFrame`. It will be less precise but as close as can be to a browser frame:

``` javascript
class FrameInterval {
    constructor(
        callback = () => {}, 
        interval = 0
    ) {
        this.callback = callback
        this.click(interval)
    }

    async click(interval, start = performance.now()) {
        while(performance.now() - start < interval){
            await new Promise(requestAnimationFrame)
        }
        this.callback()
        this.click(interval, start + interval)
    }
}

let last = performance.now()
const a = new FrameInterval(() => {
    const current = performance.now()
    console.log(current - last)
    last = current
}, 500)
```

If you are more concerned w/ the average interval being correct, you should use a `setTimeout` and adjust the delay on each iteration:

``` javascript
class DriftCompensatedInterval {
    constructor(
        callback = () => {}, 
        interval = 0
    ) {
        this.callback = callback
        this.interval = interval
        this.click()
    }

    async click(drift = 0) {
        const start = performance.now()
        setTimeout(() => {
            this.callback()
            this.click((performance.now() - start) - this.interval)
        }, this.interval - drift)
    }
}

let last = performance.now()
const a = new DriftCompensatedInterval(() => {
    const current = performance.now()
    console.log(current - last)
    last = current
}, 500)
```

nandanvasudevan:
I didn't know about any of these. Thanks for pointing it out.
I will start working on the fix.

PR #355, opened 2020-10-02
Cipher keyword shifted alphabet

bambusekd:
Adding new Cipher > Keyword shifted alphabet


raklaptudirm (commented):
A better Ceaser's cipher.

PR #555, opened 2021-01-06
Node Neighbors

rodigu:

### **Describe your change:**

* [v] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [v] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md).
* [v] This pull request is all my own work -- I have not plagiarized.
* [v] I know that pull requests will not be merged if they fail the automated tests.
* [v] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [v] All new JavaScript files are placed inside an existing directory.
* [v] All filenames should use the UpperCamelCase (PascalCase) style.  There should be no spaces in filenames.
* [v] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [v] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

### DIFF
--- a/Graphs/NodeNeighbors.js
+++ b/Graphs/NodeNeighbors.js
@@ -1,3 +1,5 @@
+// https://en.wikipedia.org/wiki/Neighbourhood_(graph_theory)
+
 class Graph{
   // Generic graph: the algorithm works regardless of direction or weight
   constructor(){
--- a/Graphs/NodeNeighbors.js
+++ b/Graphs/NodeNeighbors.js
@@ -16,7 +16,7 @@ class Graph {
 
   nodeNeighbors (node) {
     // Returns an array with all of the node neighbors
-    const neighbors = []
+    const neighbors = new Set()
     for (let i = 0; i < this.edges.length; i++) {
       // Checks if they have an edge between them and if the neighbor is not
       // already in the neighbors array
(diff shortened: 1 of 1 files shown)


Graphs/NodeNeighbors.js
```diff
@@ -1,29 +1,28 @@
...
       node1,
       node2
     })
   }
 
-  nodeNeighbors(node){
+  nodeNeighbors (node) {
     // Returns an array with all of the node neighbors
-    let neighbors = []
+    const neighbors = []
```
ruppysuppy, line 19:
```Suggestion
    const neighbors = new Set()
```
Using a `set` would reduce the time complexity from O(n ^ 2) to O(n). Use the for-of loop preferably to iterate over the edges.

ruppysuppy (changes requested):

ruppysuppy:
update the loop body & the associated function used

Graphs/NodeNeighbors.js
rodigu: Hello! 
This is my first commit to an open source project here on github
Why are sets faster in this case?
I was using lists instead of sets because of [this](https://stackoverflow.com/questions/2831212/python-sets-vs-lists), since I am iterating over every node, but the answer I read is from 10 years ago, so not really the best source.
Thanks for the feedback, really appreciate it:)

ruppysuppy: You are not iterating through it. You are checking if the neighbor is present in the list

PR #562, opened 2021-02-02
maths sysmbols list added

khantanveer7:

### **Change:**

* [v] Add an Maths Symbol List .md file


marsonya (commented):
Thanks for you contribution.
Unfortunately, this list of math symbols not what we are cataloging here. Algos and Data Structures are always welcome.
Also, the markdown is not formatted.

PR #579, opened 2021-03-17
fixed the merge implementation in MergeSort

guyariely:
In order to achieve n*log(n) time complexity, MergeSort relies on the merge function running in linear time.
Currently it runs on n^2 because of the use of the shift function.

### DIFF
--- a/Sorts/MergeSort.js
+++ b/Sorts/MergeSort.js
@@ -34,17 +34,20 @@
 *  [ ]
 */
 
-function merge (list1, list2) {
-  var results = []
+function merge(list1, list2) {
+  results = Array(list1.length + list2.length);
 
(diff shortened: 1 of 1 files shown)


Sorts/MergeSort.js
```diff
@@ -34,17 +34,20 @@
 *  [ ]
 */
 
-function merge (list1, list2) {
-  var results = []
+function merge(list1, list2) {
+  results = Array(list1.length + list2.length);
```
ruppysuppy, line 38:
```diff
- results = Array(list1.length + list2.length);
+ const results = Array(list1.length + list2.length);
```

Make sure to use `standard` to format the file as well

ruppysuppy (changes requested):

PR #618, opened 2021-05-30
Fixed the missing arguments in the countingSort() function call

meet-cv:


raklaptudirm:
@meet-cv Please fix `standardjs` style problems.

Issue #637, opened 2021-07-19
A modification to the BinarySearchTree.js


martisss:
Path：[Data-Structures/Tree/BinarySearchTree.js](https://github.com/TheAlgorithms/Javascript/blob/2d0da4d17dbd08a7cba3b70897417c252c0c7e12/Data-Structures/Tree/BinarySearchTree.js)

line 50 should be : `if (n.value <= this.value) {`,  otherwise you can't store duplicate numbers

raklaptudirm:
If you want the change please open a pull request. We will review it.

PR #639, opened 2021-07-19
modification to BinarySearchTree.js

martisss:
fixes: #637 

* [ ] modification to BinarySearchTree.js

### DIFF
--- a/Data-Structures/Tree/BinarySearchTree.js
+++ b/Data-Structures/Tree/BinarySearchTree.js
@@ -47,7 +47,7 @@ const Node = (function () {
 
   // Add a node
   Node.prototype.addNode = function (n) {
-    if (n.value < this.value) {
+    if (n.value <= this.value) {
       if (this.left == null) {
         this.left = n
       } else {


ruppysuppy (changes requested):
This might end up increasing the height of the tree unnecessarily. Instead, create a separate else block (case for the value is equal). If either of the children is null add the node, else call the function recursively on one of the children.

itsvinayak:
hey @ruppysuppy, 
I don't understand your approach quite clearly, can you please elaborate.

Thank you,
@itsvinayak

ruppysuppy:
@itsvinayak there is no point traversing down the left child if the right child is empty. The current implementation doesn't store values that are already in the tree because of `else if (n.value > this.value)`. I am suggesting something like this:

```js
Node.prototype.addNode = function (n) {
  if (n.value < this.value) {
    if (this.left == null) {
      this.left = n
    } else {
      this.left.addNode(n)
    }
  } else if (n.value > this.value) {
    if (this.right == null) {
      this.right = n
    } else {
      this.right.addNode(n)
    }
  } else {
    if (!this.left) {
      this.left = n
      return
    } else if (!this.right) {
      this.right= n
      return
    } else {
      this.left.addNode(n)
    }
}
```

PR #648, opened 2021-08-06
Added the correct(stable) version of count sort. Changed function …

mandy8055:
…declaration to ES6 style.

### **Describe your change:**

* [ ] Add an algorithm?
* [v ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**
* [v ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md).
* [v ] This pull request is all my own work -- I have not plagiarized.
* [v ] I know that pull requests will not be merged if they fail the automated tests.
* [v ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ v] All new JavaScript files are placed inside an existing directory.
* [v ] All filenames should use the UpperCamelCase (PascalCase) style.  There should be no spaces in filenames.
* [v ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [v ] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

### DIFF
--- a/Sorts/CountingSort.js
+++ b/Sorts/CountingSort.js
@@ -8,31 +8,38 @@
  * Animated Visual: https://www.cs.usfca.edu/~galles/visualization/CountingSort.html
  */
 
-function countingSort (arr, min, max) {
-  let i
-  let z = 0
+const countingSort = (arr, min, max) => {
+  // Create an auxiliary resultant array
+  const res = []
+  // Create the freq array
   const count = []
+  let len = max - min + 1
 
-  for (i = min; i <= max; i++) {
+  for (let i = 0; i < len; i++) {
     count[i] = 0
   }
-
-  for (i = 0; i < arr.length; i++) {
-    count[arr[i]]++
+  // Populate the freq array
+  for (let i = 0; i < arr.length; i++) {
+    count[arr[i] - min] += 1
   }
-
-  for (i = min; i <= max; i++) {
-    while (count[i]-- > 0) {
-      arr[z++] = i
-    }
+  // Create a prefix sum array out of the freq array
+  count[0] -= 1
+  for (let i = 1; i < count.length; i++) {
+    count[i] += count[i - 1]
   }
-
+  // Populate the result array using the prefix sum array
+  for (let i = arr.length - 1; i >= 0; i--) {
+    res[count[arr[i] - min]] = arr[i]
+    count[arr[i] - min]--
+  }
+  // Copy the result back to back to original array
+  arr = [...res]
   return arr
 }
 
 /**
-* Implementation of Counting Sort
-*/
+ * Implementation of Counting Sort
+ */
 const array = [3, 0, 2, 5, 4, 1]
 // Before Sort
 console.log('\n- Before Sort | Implementation of Counting Sort -')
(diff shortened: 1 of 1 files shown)


Sorts/CountingSort.js
```diff
@@ -8,31 +8,38 @@
...
-  let z = 0
+const countingSort = (arr, min, max) => {
+  // Create an auxiliary resultant array
+  const res = []
+  // Create the freq array
   const count = []
+  let len = max - min + 1
 
-  for (i = min; i <= max; i++) {
+  for (let i = 0; i < len; i++) {
```
raklaptudirm, line 18: Instead of doing the declaration and then the filling in the variable `count`, you can do:
```js
const count = (new Array(len)).fill(0)
```
If you do not understand how these work, refer to the following links:
1. [`Array()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array)
2. [`Array.prototype.fill()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill)

Sorts/CountingSort.js
```diff
@@ -8,31 +8,38 @@
...
+  // Populate the freq array
+  for (let i = 0; i < arr.length; i++) {
+    count[arr[i] - min] += 1
   }
-
-  for (i = min; i <= max; i++) {
-    while (count[i]-- > 0) {
-      arr[z++] = i
-    }
+  // Create a prefix sum array out of the freq array
```
raklaptudirm, line 25: All of the later code can be simplified into a few lines:
```js
for (const i in count)
  res.concat((new Array(count[i])).fill(i + min))

return res

raklaptudirm (changes requested):

Sorts/CountingSort.js
mandy8055: Hi @raklaptudirm, Thanks for reviewing. This is exactly what I wanted.

Sorts/CountingSort.js
mandy8055: In this change, I have a problem. 

- The change which you mentioned does not ensure the stability of count sort. Consider for instance, we have 1000 students. There marks are in the range 0 to 100. Now we have the sort the students based on marks. If two or more students have same marks then there roll number order should follow. But the change which you mentioned will not guarantee this. That is the main point of my PR, ensuring stability to count sort. 

- The second fix which I can think of is, it is always a good practice to return the original array where the sorting takes place. I know it won't makes difference, but generally if we are implementing an outplace algorithms, it is generally recommended to copy the contents of auxiliary array to the original array. That is the reason I didn't returned the `res` array, instead I copied the contents of `res` array to original input array `arr`.

Please do let me know If I was not clear.

Reference: [Stability of count sort](https://en.wikipedia.org/wiki/Counting_sort#Input_and_output_assumptions)

mandy8055: Hi @raklaptudirm,
I updated the code with the changes mentioned by you. Please review the commit and merge the PR.
Thanks

mandy8055 (commented):
Updated the frequency array creation and initialization. Removed the copying of result to the original array.

raklaptudirm (approved):

Issue #720, opened 2021-10-03
Consolidating some things


defaude:
Hey everyone! I just arrived here at this repo and would love to contribute a bit :)

Looking at the code organisation and the test setup, it's rather confusing for newcomers right now. Should we discuss those cross-cutting things a bit and then decide the direction going forward? Here's a few ideas - feedback very much welcome:

**Code structure**: Maybe we could define a "template" for code/algorithm files that new code should follow and/or existing code could be migrated towards. Some things to keep in mind, though:
- Implementation files (modules?) should adhere to a common structure (e.g. ESM)
- Implementation code should be stateless (i.e. it should be possible to call the implementation multiple times without side effects between test runs)
- Algorithm implementation should refrain from outputting to the console
- ... (ideas welcome)

**Test setup**: There should be just _one_ solution for testing. Do we want to use doctest going forward? Or do we want to use Jest? But keeping both is a source of confusion and will eventually lead to degrading code quality / clarity. (This is basically the attempt to revive the discussion in #142 even though I'd prefer and strongly recommend Jest)

**Chores**: Automatically running tests and code style checks in the CI is a bit _funky_ right now ;) Mostly due to the fact that there's both Jest and doctest in the mix, of course. However, there's more to do: npm dependencies are outdated, Node.js will have v16 as LTS version starting this October, etc. etc. etc.

=> What do you think? Are there things I've missed? Or do you have questions?

trasherdk:
Maybe cleaning out unused dependencies from `package.json` ?
As this thing does not have or need a runtime environment, all used dependencies should be in `devDependencies` ?
Maybe adding `husky` and a pre-commit hook, running `standard --fix .` could lower the noise ?

defaude:
That sounds like a nice idea! 👍  Maybe we could even get rid of the job that auto-updates the directory markdown file on a pre-commit basis? (hint, hint: I just improved that one with 2dc8ead...5283a41 but maybe it's even nicer to have it run as a pre-commit hook)

raklaptudirm:
@cclauss

cclauss:
An example of this _template_ would be good to see and would spark deeper conversation.  Some of these ideas could enhance the notions expressed in...
https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md#what-is-an-algorithm

raklaptudirm:
> Implementation files (modules?) should adhere to a common structure (e.g. ESM

[...]

Issue #586, opened 2021-04-14
Fix the TODOs in .github/workflows/nodejs.yml


cclauss:
https://github.com/TheAlgorithms/Javascript/blob/master/.github/workflows/nodejs.yml

raklaptudirm:
```
# TODO: error: Line 1: Unexpected token >>
```
We need to add doctests to all files to fix that issue.

cclauss:
Or only run doctest on the files that have doctests.

raklaptudirm:
Yes but I could not find any exclusion mechanism in the docs (like `/* doctest exclude */`). We can selectively run test is the workflow, but that means it has to be updated while adding any new algorithm. We need a better alternative.

cclauss:
npx doctest a/b.js c/d.js e/f/g.js

kkanekii:
Pull request please

raklaptudirm:
@kkanekii If you want to help, you can create a Pull Request.

krishdevdb:
Can someone guide me on creating a pull request?

krishdevdb:
I have to change this, right?

```diff
name: Node CI
on: [push, pull_request]
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [14.x]
    steps:
    - uses: actions/checkout@v2
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v2
      with:
        node-version: ${{ matrix.node-version }}
    - name: npm install, build, and test
      run: |
        npm install doctest standard --save-dev
-       npx doctest **/*.js || true  # TODO: Add all doctests
+      npx doctest a/b.js c/d.js e/f/g.js h/*.js
        npx standard
        npm ci
        npm run build --if-present
        npm test
      env:
        CI: true
```

cclauss:
Looks like you at least need another space.

What happens if you leave the command as it is but remove the `|| true?`?

defaude:
This will get resolved once we got rid of doctest altogether. See #742

PR #727, opened 2021-10-03
Proposal: Clean up CI job a bit

defaude:
relates to #586 and #720

Added npm scripts for doctest and style checking via standard. This allows us to call those directly via npm and not via npx.

The CI job itself is now split into distinct steps (makes it more visible which step failed).


defaude:
Sweet! The "Continuous Integration" job [looks quite nice now](https://github.com/TheAlgorithms/Javascript/pull/727/checks?check_run_id=3784372350), doesn't it?

defaude:
The job that's updating the DIRECTORY.md file has been cleaned up quite a bit, too. It's now faster and only creates a commit when there's actually a change.

raklaptudirm:
Could you fix the emerged issues?

defaude:
I did not _really_ introduce a new bug - I just applied `standard` to the code and it switched from a double-quoted string to a single-quoted string. The error was there all along, but a fix is done quickly - coming right up :)

raklaptudirm:
@cclauss knows better about these files than me, please wait for their review.

cclauss (approved):
RSLGTM

raklaptudirm:
@defaude Please resolve the conflicts so that it can be merged.

cclauss:
This job OVERWRITES DIRECTORY.md.

defaude:
The only conflict is the auto-generated DIRECTORY.md - and yes, this job deliberately overwrites this file (and has been doing so for quite some time before my refactoring, after all) ;)

PR #745, opened 2021-10-05
Project Euler - Problem 16

lvlte:

### **Describe your change:**

* [x ] Add an algorithm?
* [ ] Documentation change?

### **Checklist:**
* [x ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md).
* [x ] This pull request is all my own work -- I have not plagiarized.
* [x ] I know that pull requests will not be merged if they fail the automated tests.
* [ ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ ] All new JavaScript files are placed inside an existing directory.
* [ ] All filenames should use the UpperCamelCase (PascalCase) style.  There should be no spaces in filenames.
* [ ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.

### DIFF
--- a/Project-Euler/Problem016.js
+++ b/Project-Euler/Problem016.js
@@ -1,0 +1,45 @@
+/**
+ * Problem 16 - Power digit sum
+ *
+ * @see {@link https://projecteuler.net/problem=16}
+ *
+ * 2¹⁵ = 32768 and the sum of its digits is 3 + 2 + 7 + 6 + 8 = 26.
+ *
+ * What is the sum of the digits of the number 2¹⁰⁰⁰ ?
+ */
+
+/**
+ * Returns the power digit sum of n^pow.
+ *
+ * @param {number} [n=2]
+ * @param {number} [pow=1000]
+ * @returns {number}
+ */
+const powerDigitSum = function (n = 2, pow = 1000) {
+  // The idea is to consider each digit (d*10^exp) separately, right-to-left.
+  // digits = [units, tens, ...]
+
+  const digits = [n]
+  let p = 1
+
+  while (++p <= pow) {
+    let carry = 0
+    for (let exp = 0; exp < digits.length; exp++) {
+      const prod = digits[exp] * n + carry
+      carry = Math.floor(prod / 10)
+      digits[exp] = prod % 10
+    }
+    while (carry > 0) {
+      digits.push(carry % 10)
+      carry = Math.floor(carry / 10)
+    }
+  }
+
+  // (digits are reversed but we only want the sum so it doesn't matter)
+
+  return digits.reduce((prev, current) => prev + current, 0)
+}
+
+console.log('Power digit sum of 2^1000 :', powerDigitSum())
+
+module.exports = powerDigitSum
--- a/Project-Euler/test/Problem016.test.js
+++ b/Project-Euler/test/Problem016.test.js
@@ -1,0 +1,16 @@
+const powerDigitSum = require('../Problem016')
+
+describe('Check Problem 16 - Power digit sum', () => {
+  it('Power digit sum of 2^15', () => {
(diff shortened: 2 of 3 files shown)


Project-Euler/Problem016.js
```diff
@@ -0,0 +1,45 @@
...
+      carry = Math.floor(carry / 10)
+    }
+  }
+
+  // (digits are reversed but we only want the sum so it doesn't matter)
+
+  return digits.reduce((prev, current) => prev + current, 0)
+}
+
+console.log('Power digit sum of 2^1000 :', powerDigitSum())
```
raklaptudirm, line 43: Remove the console log.

Project-Euler/Problem016.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = powerDigitSum
```
raklaptudirm, line 45:
```suggestion
import { powerDigitSum }
```
Use ESM.

Project-Euler/test/Problem016.test.js
```diff
@@ -0,0 +1,16 @@
+const powerDigitSum = require('../Problem016')
```
raklaptudirm, line 1:
```suggestion
import { powerDigitSum } from '../Problem016'
```
Use ESM.

raklaptudirm (changes requested):

Project-Euler/Problem016.js
lvlte: I don't understand. If I remove the log then running `node Problem016.js` won't output anything, right ? 

Also, this won't be consistent with what is already comitted (algorithms available on this repo do use console.log). 

I think most people do want to see the output of an algorithm, especially when the input/arguments must fit a specific problem to be solved. But maybe I missed something ?

raklaptudirm: Algorithms are not supposed to log anything. They are supposed to be completely self contained and return all the results, not log them.

Project-Euler/Problem016.js
raklaptudirm: All the other algorithms use ESM, so there is not supposed to be a syntax error.

Project-Euler/Problem016.js
lvlte: I agree with that, and that is why by the way the console.log() is outside the exported function that describe and run the "algorithm". 

So, yes of course I can remove it, but in this case :
- Why are there all these console.log() everywhere on that repo already ?
- And why is there a `Please use 'console.log()'` put in evidence in CONTRIBUTING.md ?

Also, I insist that Project Euer problems has specific input and an expected output. It is not just an algorithm like sorting that has nothing to do/log without context/inputs.

defaude:
Hi folks - I've just now stumbled upon this discussion. @lvlte, I do understand the frustration and am sorry if there's a bit confusion going on right now.

TBH there have been several different styles how algorithm implementation code is supposed to look and how it's supposed to be run (i.e. tested):

- Simple JS files containing the algorithm itself + "test driver" code that outputs the results to the cconsole. Those algorithms are tested by simply running the whole JS file in node (CommonJS style.)
- ~~JS files with doctest comments~~ (we just got rid of those)
- JS files that adhere to ESM + dedicated test files. Tests are run through Jest, which picks up the babel config.
- ... and a bit of a wild mix altogether ;)

We discussed this in #720 where we decided that, the code shall be more consistent. Going forward, it makes sense to push towards ESM only (eventually just add the `"type": "module"` property in the package.json and remove the need for babel altogether. But we're not there yet ;)

So please try to write new additions ESM-style, without console.log, with `export function myAlgorithm` etc. + tests written in Jest. While this may be inconsistent _at this moment_ with previous contributions, we're pushing towards migrating all existing code into the targeted style.

PR #768, opened 2021-10-09
Changes for consolidation

### LINKED ISSUE #720: Consolidating some things

lvlte:
Full switch to ESM. 
Remove live code examples (either convert to Jest test or leave hints as a comment). 
Remove all console.log (except if encapsulated within a specific method/function).
Revise all algorithms logging results instead of returning them. 
...

Fixes #720

### DIFF
--- a/Conversions/RomanToDecimal.js
+++ b/Conversions/RomanToDecimal.js
@@ -8,7 +8,7 @@ const values = {
   M: 1000
 }
 
-function romanToDecimal (romanNumber) {
+export function romanToDecimal (romanNumber) {
   let prev = ' '
 
   let sum = 0
@@ -32,5 +32,3 @@ function romanToDecimal (romanNumber) {
   }
   return sum
 }
-
-export { romanToDecimal }
--- a/Conversions/test/RomanToDecimal.test.js
+++ b/Conversions/test/RomanToDecimal.test.js
@@ -1,4 +1,4 @@
-import { romanToDecimal } from '../romanToDecimal'
+import { romanToDecimal } from '../RomanToDecimal'
 
 describe('romanToDecimal', () => {
   it('XXIIVV', () => {


raklaptudirm:
Is this done? If it is not, please convert the pr to draft.

raklaptudirm:
@lvlte Please try to avoid using the jest mock functions. Remove the console logs wherever possible, and only use the mock functions where the console logs are a part of the solution.

lvlte:
It was tedious but I think I'm finally done with this PR.

lvlte:
Thanks, I've added the keyword in the PR description.

PR #785, opened 2021-10-13
Pow function optimization using exponentiation by squaring

farha1:

### **Describe your change:**

* [ ] Add an algorithm?
* [x] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/Maths/Pow.js
+++ b/Maths/Pow.js
@@ -1,18 +1,11 @@
 // Returns the value of x to the power of y
 
 const pow = (x, y) => {
-  let result = 1
-  if (y >= 0) {
-    for (let i = 1; i <= y; i++) {
-      result *= x
-    }
-  } else {
-    for (let i = y; i < 0; i++) {
-      result *= x
-    }
-    result = 1 / result
-  }
-  return result
+  if (y < 0) return pow(1 / x, -y)
+  else if (y === 0) return 1
+  else if (y === 1) return x
+  else if (y % 2 === 0) return pow(x * x, y / 2)
+  else if (y % 2 !== 0) return x * pow(x * x, (y - 1) / 2)
 }
 
 export { pow }
--- a/Maths/Pow.js
+++ b/Maths/Pow.js
@@ -2,8 +2,15 @@
 
 const pow = (x, y) => {
   let result = 1
-  for (let i = 1; i <= y; i++) {
-    result *= x
+  if (y >= 0) {
+    for (let i = 1; i <= y; i++) {
+      result *= x
+    }
+  } else {
+    for (let i = y; i < 0; i++) {
+      result *= x
+    }
+    result = 1 / result
   }
   return result
 }
--- a/Maths/Pow.js
+++ b/Maths/Pow.js
@@ -1,4 +1,5 @@
 // Returns the value of x to the power of y
+// Exponentiation by squaring : https://en.wikipedia.org/wiki/Exponentiation_by_squaring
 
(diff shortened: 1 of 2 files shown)


tjgurwara99 (commented):
Isn't this slower? This is using a recursive approach - arguably slower than the iterative one. You should add a separate function instead of overwriting the already fast implementation. If you still think this is fast, could you write a benchmark to see the differences.

Also, note that I'm not against having your approach - in fact, your approach is more mathematical but I question that the title of this PR is suitable for what you're trying to do here.

booleans-oss:
I've decided to try and benchmark this new implementation. The screenshots provided are only tested with the number 5, 10, 20 and 50. And the exponents go from 0 to 100. Larger variables could be used but I noticed that the pattern stay approximately the same so I decided to stay in that range.

<details>
<summary>Screenshots</summary>

time-5

time-10

time-20

time-50

</details>

In my opinion, the difference in execution between the two implementations is too small to categorically say that one is faster than the other. We can clearly see some peaks but not significantly enough to choose between the two.

We also should keep in mind that the code that I used to measure the elapsed time might not be perfect. There might be a lot more variables affecting the execution time. (If needed, I can provide the benchmarking code I made).  

I hope it helps.

booleans-oss:
I made a simple benchmarking workspace to see the execution time difference for the two implementations. For simplicity sake, I use any integers from 0 to 100 and exponents range from 0 to 10000. 

<details>
<summary>Screenshots</summary>

time-10

time-100

time-1000

time-10000

</details>

The screenshots show the same results as the previous benchmarks: **the previous implementation is slower than the new one**.

raklaptudirm:
The benchmarks are good, but I would still suggest to keep both implementations.

tjgurwara99 (changes requested):
It's good that you improved this, however my previous comment was just suggesting that you shouldn't remove the previous implementation. I still think you should NOT REMOVE the old implementation but rather create a separate function, calling `powIterative` and `powRecursive` or something along those lines.

raklaptudirm:
@farha1 seems like you abandoned this pr. Please go through with the review comments if you still want to work on this.

appgurueu:
> Isn't this slower? This is using a recursive approach - arguably slower than the iterative one. You should add a separate function instead of overwriting the already fast implementation. If you still think this is fast, could you write a benchmark to see the differences.

No, this implementation is way faster in fact. The naive algorithm does O(n) multiplications where n is the exponent. This algorithm halves the exponent in each step. The recursive call tree thus has logarithmic depth. Yet the effort per call is constant. So this is O(log n), way faster.

In practice the iterative approach might be faster for reasonably small exponents.

I too suggest keeping both, but this is generally the preferable approach.

Suggestion to make it faster: You don't need to check for all edge cases (0, 1, negative exponents) if you check for a couple in the calling func already and then call a recursive helper which only receives the exponents you expect.

appgurueu (approved):

raklaptudirm:
@farha1 You need to fix the branch conflicts.

tjgurwara99:
> No, this implementation is way faster in fact. The naive algorithm does O(n) multiplications where n is the exponent. This algorithm halves the exponent in each step.

Yeah, didn't notice the halving of the exponent😅

Though still think instead of overwriting the function, we should keep both.

raklaptudirm:
@farha1 seems to have abandoned this pr.

PR #787, opened 2021-10-14
Decimal Expansion

lvlte:
Adding algorithm for representing the decimal expansion of a given fraction a/b using euclidean division. Optionally, the representation can be in binary, ternary, octal, any base from 2 to 10. 

Code is compliant with Javascript standard and covered by Jest tests.


lvlte:
In this particular case, I prefer array vs object, because it left the freedom to name variables when doing destructuring assignments, which is probably better here because this function can return fraction decimals by default, but also bits, trits, octals, etc.

Using an object, variables identitifers will always be restricted to the object properties, regardless of the base, which might be more conducive to confusion if the base used is not 10 : 
`const {integer, decimal, indexOf} = decExp(a, b, 8);`

Using an array leaves more freedom : 
`const [integer, decimals, index] = decExp(a, b);`
`const [integer, octals, index] = decExp(a, b, 8);`
`const [int, trits, i] = decExp(a, b, 3)`

This is like hard vs soft code for variable naming/assignment, it encourages the developer to define (meaningful) variables by himself, depending on the situation (btw I realized I refer to decimals regardless of the base in the tests). But if he does not, I agree it is not an advantage at all, eg. in this case :

```
decExpArray = decExp(a, b);
// doing things with decExpArray[0]
// doing things with decExpArray[1]]
// doing things with decExpArray[2] ... 
```

What's your opinion ?

PR #846, opened 2021-11-24
fix #844

### LINKED ISSUE #844: LRU Cache js code can be more simple and optimized using builtin Map data structure
Map stores order of insertion and internally use a doubly-linked list to achieve this.

sample code to demonstrate: get and put time complexity: O(1)

```
/**
 * @param {number} capacity
 */
var LRUCache = function(capacity) {
    this.capacity = capacity;
    this.cache = new Map();
};

/** 
 * @param {number} key
 * @return {number}
 */
LRUCache.prototype.get = function(key) {
    if(this.cache.has(key)){
            let value = this.cache.get(key);
            this.cache.delete(key);
            this.cache.set(key,value);
            return value;
        }
        else{
            return -1;
        }
};

/** 
 * @param {number} key 
 * @param {number} value
 * @return {void}
 */
LRUCache.prototype.put = function(key, val) {
    if(this.cache.has(key)){
            //refresh the cache
            this.cache.delete(key);
            this.cache.set(key,val);
        }
        else{
            if(this.cache.size===this.capacity){
                this.cache.delete(this.cache.keys().next().value);
            }
            this.cache.set(key,val);
        }
};
```

aditya1944:
Optimized code for LRU Cache.

How?
Used order of insertion functionality in the inbuilt data structure for the map instead of creating a custom doubly linked list.

### DIFF
--- a/Cache/LRUCache.js
+++ b/Cache/LRUCache.js
@@ -1,88 +1,39 @@
-class DoubleLinkedListNode {
-  // Double Linked List Node built specifically for LRU Cache
-  constructor (key, val) {
-    this.key = key
-    this.val = val
-    this.next = null
-    this.prev = null
-  }
-}
-
-class DoubleLinkedList {
-  // Double Linked List built specifically for LRU Cache
-  constructor () {
-    this.head = new DoubleLinkedListNode(null, null)
-    this.rear = new DoubleLinkedListNode(null, null)
-    this.head.next = this.rear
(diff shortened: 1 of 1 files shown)


raklaptudirm (changes requested):
Format your code with `standard.js`.

aditya1944:
@raklaptudirm Code is formatted using `standard.js`.

raklaptudirm (approved):

PR #868, opened 2021-12-09
improve average

RuSaG0:
Fix a bug(check divide by zero) and improve of calc (can work with generators and custom selectors)

### DIFF
--- a/Maths/AverageMean.js
+++ b/Maths/AverageMean.js
@@ -1,29 +1,26 @@
-'use strict'
 /*
-    author: PatOnTheBack
-    license: GPL-3.0 or later
-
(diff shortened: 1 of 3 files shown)


raklaptudirm (commented):
As I mentioned, not need to support all iterables. A simple array length check should suffice.

raklaptudirm:
@RuSaG0 As we discussed in #870, implement that method as a separate function.

appgurueu:
> @RuSaG0 As we discussed in #870, implement that method as a separate function.

I agree, this is not really an improvement over the current code. Please open a new PR if you plan on adding this; this one has merge conflicts anyways.

PR #869, opened 2021-12-09
improve median

RuSaG0:
### Describe your change:

Fix a bug
In the last implementation, **finding the median is not a pure function**. It mutates the original array.

const arr = [3,2,1]
averageMedian(arr)
`numbers = numbers.sort(sortNumbers)`
so arr will be [1,2,3].

Why? The function for finding the median should not change arr by sorting

====
Now this fn can work with generators and custom median selector (for example of str lengths)


raklaptudirm (commented):
Any iterables can be converted to an array by simply calling `Array.from(iterator)`. I believe the previous implementation was more concise and to the point.

If you want a pure function, I would suggest a "rest" argument for the function.

RuSaG0:
>>>  I believe the previous implementation was more concise and to the point. - no problem 
I returned the previous one, only now the passed array doesn't change

===
updated

Maths/AverageMedian.js
```diff
@@ -1,14 +1,5 @@
...
- * function averageMedian
- * to find the median value of an array of numbers
- * the numbers in an array will be sorted in ascending order by the function sortNumbers
- * if the length of the array is even number, the median value will be the average of the two middle numbers
- * else if the length of the array is odd number, the median value will be the middle number in the array
- */
-
-const averageMedian = (numbers) => {
+const averageMedian = (_numbers) => {
+  let numbers = [..._numbers];
```
raklaptudirm, line 2: I don't understand why any of the changes other than this line is required. Keep this line and revert all the other changes.

raklaptudirm (changes requested):

Maths/AverageMedian.js
```diff
@@ -8,7 +8,8 @@
  * else if the length of the array is odd number, the median value will be the middle number in the array
  */
 
-const averageMedian = (numbers) => {
+const averageMedian = (_numbers) => {
```
raklaptudirm, line 11: This repo requires all variable names to be in lowerCamelCase and without underscores. Use different names for the variables.

Maths/AverageMedian.js
```diff
@@ -26,4 +27,4 @@ const sortNumbers = (num1, num2) => {
   return num1 - num2
 }
 
-export { averageMedian }
+export { averageMedian }
```
raklaptudirm, line 30: All files should have an ending newline, do not remove it.

Maths/test/AverageMedian.test.js
```diff
@@ -18,4 +18,4 @@ test('should return the median of an array of numbers:', () => {
 test('should return the median of an array of numbers:', () => {
   const medianValue = averageMedian([1, 2, 3, 4, 6, 8])
   expect(medianValue).toBe(3.5)
-})
+})
```
raklaptudirm, line 21: Dont remove ending newline.

raklaptudirm (changes requested):

Maths/test/AverageMedian.test.js
RuSaG0: yeah, thx, i forget about ur CONTRIBUTING.md

one indication would be enough)

Maths/AverageMedian.js
RuSaG0: and without underscores - But why?

I may not have so much commercial experience, but wherever I have been, they used it.
This immediately makes it possible to distinguish input parameters from local variables and errors, like mutation input data, would most likely not occur

RuSaG0: Also easy work with autocomplete from ide. 
It almost never suggests incorrect values when autocomplete , unlike the suggested option.

Maths/test/AverageMedian.test.js
RuSaG0: total: 10 comments and 4 commits for two lines of code)))

Maths/AverageMedian.js
raklaptudirm: It is just a style followed by this repository, to make variable naming consistent. Also, as I previously mentioned, these algorithms do not need to be production ready of useful, just interesting and educative from a programming perspective.

Maths/test/AverageMedian.test.js
raklaptudirm: Those two lines of code were the crux of what you were trying to achieve. All the other code was to try and generalize this method, which is not required in this context. But even small contributions are useful in making this repository better.

raklaptudirm (approved):

PR #870, opened 2021-12-09
findmin

RuSaG0:
improve find Min method
if no params to calc min => it's not error, it's undefined/null
work with generators, custom selector fn

### DIFF
--- a/Maths/FindMin.js
+++ b/Maths/FindMin.js
@@ -1,23 +1,41 @@
 /**
  * @function FindMin
  * @description Function to find the minimum number given in an array of integers.
- * @param {Integer[]} nums - Array of Integers
- * @return {Integer} - The minimum number of the array.
  */
 
-const findMin = (...nums) => {
-  if (nums.length === 0) {
-    throw new TypeError('Array is empty')
-  }
+const findMin = (_iterable, _selector = undefined) => {
+  let min = undefined;
 
-  let min = nums[0]
-  for (let i = 1; i < nums.length; i++) {
-    if (nums[i] < min) {
-      min = nums[i]
+  const iterator = _iterable[Symbol.iterator]();
+  if (!_selector) {
+    let current = iterator.next();
+    if(current.done)
+      return undefined;
+    min = current.value;
+
+    current = iterator.next();
+    while(!current.done) {
+      const x = current.value;
+      if(x < min)
+        min = x;
+      current = iterator.next();
     }
   }
+  else {
+    let current = iterator.next();
+    if(current.done)
+      return undefined;
+    min = _selector(current.value);
 
-  return min
+    current = iterator.next();
+    while(!current.done) {
+      const x = _selector(current.value);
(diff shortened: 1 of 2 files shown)


raklaptudirm (commented):
Please explain how your implementation is better. I for one, think the previous one was the better one.

RuSaG0:
If the goal is learning and understanding what the search for the minimum is - the previous implementation is better (although, as it seems to me, everyone already understands what the search for the minimum means)

If we need a more 'production-ready function' that is able to work with different structures (iterables for example) and with different ways to search for the minimum itself, then mine is better. Finding the minimum is not always looking for the smallest number

raklaptudirm:
This repo is not for production ready code, but for self explanatory code understandable for beginners of the programming language. Throwing iterators at them is not a good idea imo.

trasherdk:
I like this implementation.

Maybe as an alternative `FindMinIterator()` or something.

```js
  test('given array of objects then min accessor is found', () => {
    const array = [
      { name: 'Item #1', price: 1.0 },
      { name: 'Item #1', price: 0.0 },
      { name: 'Item #1', price: -1.0 }
    ]
    expect(findMin(array, _x => _x.price)).toBe(-1)
  })
```

The tests pass
```sh
$ npx jest --no-cache Maths/test/FindMin.test.js 
 PASS  Maths/test/FindMin.test.js
  findMin
    ✓ given empty array then min is undefined (2 ms)
    ✓ given single value array then min is found (1 ms)
    ✓ given array then min is found
    ✓ given empty generator then min is undefined
    ✓ given generator then min is found
    ✓ given string generator then min string length is found
    ✓ given array of objects then min accessor is found (1 ms)

Test Suites: 1 passed, 1 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        1.147 s
Ran all test suites matching /Maths\/test\/FindMin.test.js/i.
```

booleans-oss:
> This repo is not for production ready code, but for self explanatory code understandable for beginners of the programming language. Throwing iterators at them is not a good idea imo.

I agree. Iterators and generators are really difficult topics to understand and even more to explain. I think it is far more easier and useful to explain a simple for-loop (as seen in the previous implementation) than explaning iterators. So in my opinion and for the target audience of this repo, I think the previous implementation is better.

raklaptudirm:
@trasherdk I completely agree on the fact that this implementation is faster and more useful than the current one, but thats not my point.

trasherdk:
@raklaptudirm I get your point. Don't complicate stuff :smiley: 

I just liked the implementation, and the use case I demonstrated.

raklaptudirm:
We could add this as a separate implementation though.

trasherdk:
That would work too :smiley: 

What do you say @RuSaG0 ?

RuSaG0:
@[trasherdk](https://github.com/trasherdk)
yes, let's make it as second method (as you say "FindMinIterator()")

Issue #883, opened 2022-02-16
Required parameter best practice, avoid nested if block for better performance and readability


AbmSourav:
**binarySearch**
* required parameters and optional parameters are ordered in a mixed manner. A common programming convention is adding required parameters at the left and optional parameters at the right.
* In recursive function, early return base case is better for performance and readability.

PR #884, opened 2022-02-16
binarySearch

AbmSourav:
* Parameters added according to the convention
* Early return base case added

#883


raklaptudirm (changes requested):
Please format your code with `standard.js`.

raklaptudirm (approved):

PR #896, opened 2022-02-20
added ES6 (specifically replaced constructor function with class syntax) version of DoublyLinkedList

mmestiyak:


raklaptudirm (changes requested):
No need to create a new implementation, replace the current one with this.

PR #900, opened 2022-02-22
replaced constructor function with es6 class syntax

mmestiyak:
# replaced constructor function with es6 class syntax


raklaptudirm (changes requested):
Format your code with `standard.js`.

mmestiyak:
Hello @raklaptudirm,  formatted code with standard.js please checkout

raklaptudirm (approved):

Issue #901, opened 2022-02-23
Anagram is or not case sensitive!


fahimfaisaal:
From my research, I've come to know anagram is not case sensitive, but this [checkAnagram](https://github.com/TheAlgorithms/Javascript/blob/master/String/CheckAnagram.js) function is case-sensitive. it's can't pass this test case **“Eleven plus two” is an anagram of “Twelve plus one”**

raklaptudirm:
Please open a pull request to fix this issue.

fahimfaisaal:
Ok

PR #902, opened 2022-02-24
Upgrade `checkAnagram` function & Fixes

fahimfaisaal:

### Describe your change:

* [x] Optimize the algorithm via `Array.prototype.reduce` & `String.prototype.replace` methods
* [x] Fix a bug case-sensitive to insensitive by using `RegExp`
* [x] Add new JS Doc
* [x] Add new test cases for invalid input & case-insensitive
* [x] **Fixed:** #901 

### Checklist:

### DIFF
--- a/String/CheckAnagram.js
+++ b/String/CheckAnagram.js
@@ -11,7 +11,7 @@
 const checkAnagramRegex = (str1, str2) => {
   // check that inputs are strings.
   if (typeof str1 !== 'string' || typeof str2 !== 'string') {
-    throw new TypeError('Arguments should be string both')
+    throw new TypeError('Both arguments should be strings.')
   }
 
   // If both strings have not same lengths then they can not be anagram.
@@ -42,7 +42,7 @@ const checkAnagramRegex = (str1, str2) => {
 const checkAnagramMap = (str1, str2) => {
   // check that inputs are strings.
   if (typeof str1 !== 'string' || typeof str2 !== 'string') {
-    throw new TypeError('Arguments should be string both')
+    throw new TypeError('Both arguments should be strings.')
   }
 
   // If both strings have not same lengths then they can not be anagram.
--- a/String/CheckAnagram.js
+++ b/String/CheckAnagram.js
@@ -1,14 +1,14 @@
 // An [Anagram](https://en.wikipedia.org/wiki/Anagram) is a string that is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once. Anagram check is not case-sensitive;
 /**
- * @function checkAnagramViaRegex
+ * @function checkAnagramRegex
  * @param {string} str1
  * @param {string} str2
  * @returns {boolean}
  * @description - check anagram with the help of Regex
- * @example - checkAnagramViaRegex('node', 'deno') => true
- * @example - checkAnagramViaRegex('Eleven plus two', 'Twelve plus one') => true
+ * @example - checkAnagramRegex('node', 'deno') => true
+ * @example - checkAnagramRegex('Eleven plus two', 'Twelve plus one') => true
  */
-const checkAnagramViaRegex = (str1, str2) => {
+const checkAnagramRegex = (str1, str2) => {
   // check that inputs are strings.
   if (typeof str1 !== 'string' || typeof str2 !== 'string') {
     throw new TypeError('Arguments should be string both')
@@ -31,15 +31,15 @@ const checkAnagramViaRegex = (str1, str2) => {
 }
 
 /**
- * @function checkAnagramViaMap
+ * @function checkAnagramMap
  * @description - check anagram via using HashMap
  * @param {string} str1
  * @param {string} str2
  * @returns {boolean}
-* @example - checkAnagramViaMap('node', 'deno') => true
- * @example - checkAnagramViaMap('Eleven plus two', 'Twelve plus one') => true
+* @example - checkAnagramMap('node', 'deno') => true
+ * @example - checkAnagramMap('Eleven plus two', 'Twelve plus one') => true
  */
-const checkAnagramViaMap = (str1, str2) => {
+const checkAnagramMap = (str1, str2) => {
   // check that inputs are strings.
   if (typeof str1 !== 'string' || typeof str2 !== 'string') {
     throw new TypeError('Arguments should be string both')
@@ -73,4 +73,4 @@ const checkAnagramViaMap = (str1, str2) => {
   return true
 }
 
-export { checkAnagramViaRegex, checkAnagramViaMap }
+export { checkAnagramRegex, checkAnagramMap }
--- a/String/CheckAnagram.js
+++ b/String/CheckAnagram.js
@@ -36,7 +36,7 @@ const checkAnagramRegex = (str1, str2) => {
  * @param {string} str1
  * @param {string} str2
  * @returns {boolean}
-* @example - checkAnagramMap('node', 'deno') => true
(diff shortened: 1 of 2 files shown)


String/CheckAnagram.js
```diff
@@ -1,47 +1,28 @@
-// An [Anagram](https://en.wikipedia.org/wiki/Anagram) is a string that is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once.
-// Anagram check is case sensitive; i.e. Aba and aba is not a anagram.
-// inputs are strings i.e. str1 and str2
+/**
+ * @function checkAnagram
+ * @param {string} str1
+ * @param {string} str2
+ * @returns {boolean}
+ * @description An [Anagram](https://en.wikipedia.org/wiki/Anagram) is a string that is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once. Anagram check is case sensitive;
```
raklaptudirm, line 6:
```suggestion
 * @description An [Anagram](https://en.wikipedia.org/wiki/Anagram) is a string that is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once. Anagram check is not case sensitive;
```

String/CheckAnagram.js
```diff
@@ -1,47 +1,28 @@
...
-
-  // Now check if all entries in hashmap has zeros.
-
-  for (const key in str1CharCount) {
-    if (str1CharCount[key] !== 0) return false
-  }
-
-  return true
+  // str1 converted to array and traverse the each letter of str1 by reduce method
+  return ![...str1].reduce(
```
raklaptudirm, line 22: This line may be confusing for people who a new to Javascript. I would suggest properly explaining the substitution and boolean type casting steps in the comments.

raklaptudirm (changes requested):

kikar:
> Optimize the algorithm via Array.prototype.reduce & String.prototype.replace methods
Looks like the algorithm is less efficient now. It was O(n) and now is O(n^2)

raklaptudirm:
@kikar I think the new implementation can teach more Javascript concepts than the previous one, which is our primary aim. That said, I think it will be better to have the two implementations separately @fahimfaisaal.

String/CheckAnagram.js
fahimfaisaal: I think so, thanks for your suggestion.

fahimfaisaal:
@kikar to my knowledge, the `String.prototype.replace` method used regex that's why the operation is very cheap, it's not traversing the full string `str2Acc` in every traverse of `Array.prototype.reduce` method. it just removes **1 char** from `str2Acc` in 1 `Array.prototype.reduce` method traverse. which char is just matched to the `cur` variable.

fahimfaisaal:
@kikar to my knowledge, the `String.prototype.replace` method used **regex** that's why the operation is very cheap, it's not traversing the full string `str2Acc` in every traverse of `Array.prototype.reduce` method. it just removes `1 char` from `str2Acc` in 1 `Array.prototype.reduce` method traverse. which char is just matched to the cur variable.

So, the algorithm couldn't be O(n^2)

fahimfaisaal:
@kikar  to my knowledge, the `String.prototype.replace` method used **regex** that's why the operation is very cheap, cause it's not traversing the full string `str2Acc` in every traverse of `Array.prototype.reduce` method. it just removes **1 char** from `str2Acc` in 1 `Array.prototype.reduce` method traverse. which char is just matched to the `cur` variable.

So, the algorithm couldn't be **O(n^2)** complexity.

kikar:
> @kikar to my knowledge, the `String.prototype.replace` method used **regex** that's why the operation is very cheap, cause it's not traversing the full string `str2Acc` in every traverse of `Array.prototype.reduce` method. it just removes **1 char** from `str2Acc` in 1 `Array.prototype.reduce` method traverse. which char is just matched to the `cur` variable.
> 
> So, the algorithm couldn't be **O(n^2)** complexity.

Even if use regex, the code needs to find the first occurrence of the letter selected. Take the case of the reversed string.
For each letter from str1 you need to traverse the whole str2 to find the letter to remove.

fahimfaisaal:
> @kikar I think the new implementation can teach more Javascript concepts than the previous one, which is our primary aim. That said, I think it will be better to have the two implementations separately @fahimfaisaal.

@raklaptudirm I think if I optimize my algorithm, It's shouldn't add two more implementations. please let me know what could I do.

String/CheckAnagram.js
fahimfaisaal: FIXED

raklaptudirm:
Since the two implementations are very different, I would suggest to keep and test both.

fahimfaisaal:
@raklaptudirm I reckon, if we add this algorithm it's will replace previous algos.
```javascript
const str1List = Array.from(str1.toUpperCase()) // str1 to array 

  // get the occurrences of characters of str1 by using HashMap
  const str1Occurs = str1List.reduce(
    (map, char) => (
      map.set(char, map.get(char) + 1 || 1), map
    ),
    new Map()
  )

  for (const char of str2.toUpperCase()) {
    // if char has not exist to the map it's return false
    if (!str1Occurs.has(char)) {
      return false;
    }

    let getCharCount = str1Occurs.get(char)
    str1Occurs.set(char, --getCharCount)

    getCharCount === 0 && str1Occurs.delete(char)
  }

  return true;
}
```

raklaptudirm:
Both do not need to have the same name.

fahimfaisaal:
Ok I will add these two with different names, but could I add these two algos in One file?

raklaptudirm:
Yes.

fahimfaisaal:
Ok, thanks for your rapid response.

String/CheckAnagram.js
```diff
@@ -18,11 +19,58 @@ const checkAnagram = (str1, str2) => {
...
+/**
+ * @function checkAnagramViaMap
+ * @description - check anagram via using HashMap
+ * @param {string} str1
+ * @param {string} str2
+ * @returns {boolean}
+* @example - checkAnagramViaMap('node', 'deno') => true
+ * @example - checkAnagramViaMap('Eleven plus two', 'Twelve plus one') => true
+ */
+const checkAnagramViaMap = (str1, str2) => {
```
raklaptudirm, line 42:
```suggestion
const checkAnagramMap = (str1, str2) => {
```

raklaptudirm (changes requested):
Rename the functions to remove `via` from the name.

String/CheckAnagram.js
```diff
@@ -1,13 +1,14 @@
...
  * @returns {boolean}
- * @description An [Anagram](https://en.wikipedia.org/wiki/Anagram) is a string that is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once. Anagram check is not case-sensitive;
- * @example - checkAnagram('node', 'deno') => true
- * @example - checkAnagram('Eleven plus two', 'Twelve plus one') => true
+ * @description - check anagram with the help of Regex
+ * @example - checkAnagramViaRegex('node', 'deno') => true
+ * @example - checkAnagramViaRegex('Eleven plus two', 'Twelve plus one') => true
  */
-const checkAnagram = (str1, str2) => {
+const checkAnagramViaRegex = (str1, str2) => {
```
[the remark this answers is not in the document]
fahimfaisaal: Done

fahimfaisaal:
You're really strict reviewer. 😀

raklaptudirm (approved):

PR #907, opened 2022-02-28
Add `CheckExceeding` function

### LINKED ISSUE #765: a new challenge alogorithm
Exceeding words are words where the gap between two adjacent characters is increasing.
The gap is the distance in ascii

Example:

input =  update
output = False 

Explanation :
from u to p: 5
from p to d: 12
from d to a: 3
from a to t: 19
from t to e: 15

5,12,3,19,15 is not increasing.
The word "update" is not an Exceeding Word

fahimfaisaal:

### Describe your change:

* [x] Add check exceeding algorithm.
* [x] Add test case for the algorithm.
* [x] Resolved #765 

### Checklist:

### DIFF
--- a/String/CheckExceeding.js
+++ b/String/CheckExceeding.js
@@ -1,0 +1,42 @@
+/**
+ * @function checkExceeding
+ * @description - Exceeding words are words where the gap between two adjacent characters is increasing. The gap is the distance in ascii
+ * @param {string} str
+ * @returns {boolean}
+ * @example - checkExceeding('delete') => true, ascii difference - [1, 7, 7, 15, 15] which is incremental
+ * @example - checkExceeding('update') => false, ascii difference - [5, 12, 3, 19, 15] which is not incremental
+ */
+const checkExceeding = (str) => {
+  if (typeof str !== 'string') {
+    throw new TypeError('Argument is not a string')
+  }
+
+  const upperChars = str
+    .toUpperCase()
+    .replace(/[^A-Z]/g, '') // remove all from str except A to Z alphabets
+
+  const adjacentDiffList = []
+
+  for (let i = 0; i < upperChars.length - 1; i++) {
+    // destructuring current char & adjacent char by index, cause in javascript String is an object.
+    const { [i]: char, [i + 1]: adjacentChar } = upperChars
+
+    if (char !== adjacentChar) {
+      adjacentDiffList.push(
+        Math.abs(char.charCodeAt() - adjacentChar.charCodeAt())
+      )
+    }
+  }
+
+  for (let i = 0; i < adjacentDiffList.length - 1; i++) {
+    const { [i]: charDiff, [i + 1]: secondCharDiff } = adjacentDiffList
+
+    if (charDiff > secondCharDiff) {
+      return false
+    }
+  }
+
+  return true
+}
+
+export { checkExceeding }
--- a/String/test/CheckExceeding.test.js
+++ b/String/test/CheckExceeding.test.js
@@ -24,7 +24,15 @@ describe('Testing CheckExceeding function', () => {
     expect(checkExceeding('Hello')).toBe(false)
     expect(checkExceeding('world')).toBe(false)
     expect(checkExceeding('update')).toBe(false)
-    expect(checkExceeding('upgrade')).toBe(false)
+    expect(checkExceeding('university')).toBe(false)
+    expect(checkExceeding('dog')).toBe(false)
+    expect(checkExceeding('exceeding')).toBe(false)
+    expect(checkExceeding('resolved')).toBe(false)
+    expect(checkExceeding('future')).toBe(false)
+    expect(checkExceeding('fixed')).toBe(false)
+    expect(checkExceeding('codes')).toBe(false)
+    expect(checkExceeding('facebook')).toBe(false)
+    expect(checkExceeding('vscode')).toBe(false)
   })
 
   it('Testing exceeding words', () => {
--- a/String/test/CheckExceeding.test.js
+++ b/String/test/CheckExceeding.test.js
@@ -1,0 +1,38 @@
+import { checkExceeding } from '../CheckExceeding'
+
+describe('Testing CheckExceeding function', () => {
+  it('Testing the invalid types', () => {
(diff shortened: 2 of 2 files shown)


String/CheckExceeding.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const adjacentDiffList = []
+
+  for (let i = 0; i < upperChars.length - 1; i++) {
+    // destructuring current char & adjacent char by index, cause in javascript String is an object.
+    const { [i]: char, [i + 1]: adjacentChar } = upperChars
+
+    if (char !== adjacentChar) {
+      adjacentDiffList.push(
+        Math.abs(char.charCodeAt() - adjacentChar.charCodeAt())
```
raklaptudirm, line 26: I don't believe they used the abs in the specification.

String/test/CheckExceeding.test.js
```diff
@@ -0,0 +1,38 @@
...
+  it('Testing not exceeding words', () => {
+    expect(checkExceeding('Hello')).toBe(false)
+    expect(checkExceeding('world')).toBe(false)
+    expect(checkExceeding('update')).toBe(false)
+    expect(checkExceeding('upgrade')).toBe(false)
+  })
+
+  it('Testing exceeding words', () => {
+    expect(checkExceeding('bee')).toBe(true) // [ 3 ]
+    expect(checkExceeding('can')).toBe(true) // [ 2, 13 ]
```
raklaptudirm, line 32: Why is `can` exceeding? c > a < n

raklaptudirm (changes requested):

String/CheckExceeding.js
fahimfaisaal: From the problems example, the increasing of "update" = `[5,12,3,19,15]` is **abs**, cause except **abs** the increasing should be `[ 5, 12, 3, -19, 15 ]`
difference between `a:97 - t:116 = -19 = |19|`

this is why I added the **abs** method.

String/test/CheckExceeding.test.js
fahimfaisaal: By the rules of the **lexicographical system**, your statement is 100% true. but this solution does not follow the lexicographical system.

String/CheckExceeding.js
raklaptudirm: You are doing the opposite. Do `t:116 - a:97 = 19`. Otherwise everything will be considered an increasing.

String/test/CheckExceeding.test.js
raklaptudirm: More context?

String/CheckExceeding.js
fahimfaisaal: If I do `t:116 - a:97 = 19` this output should be the same. I've added more **not-increasing** words please check.

String/test/CheckExceeding.test.js
fahimfaisaal: If I add a lexicographical condition, It'll be hard to find the increasing words. Only linear alphabets exist in the exceeding word list, which has no meaning and have very tiny examples.

fahimfaisaal: If I add **lexicographical** condition, it's hard to find the **increments** words.

raklaptudirm: No, I wanted to know what definition of an exceeding word are you using.

fahimfaisaal: From the problem, the definition is, every **ASCII** difference of **char** & **adjacentChar** sequence are must be increasing.

raklaptudirm: Yes, but if you do a `Math.Abs()`, all of them will be positive, so increasing, so I don't get how this works.

fahimfaisaal: I used `Math.abs()` for the "update" example that increasing `[5, 12, 3, 19, 15]`, not `[ 5, 12, 3, -19, 15 ]`

raklaptudirm: I still do not understand what you are trying to do. I will try to make sense of the algorithm again tomorrow, but in the mean type, provide an explanation of how the algorithm works.

fahimfaisaal: From the challenge 
> Exceeding words are words where the gap between two adjacent characters is increasing.
**The gap is the distance in ASCII**

Example: **"update** is not exceeding, because the ASCII distance of each pair isn't increasing.

Explanation of **update** word: 
from `u:117` to `p:112` ->  `117 - 112 = |5| = 5`
from `p:112` to `d:100` -> `112 - 100 = |12| = 12`
from `d:100` to `a:97` -> `100 - 97 = |3| = 3`
from `a:97` to `t:116`: -> `97 - 116 = |-19| = 19`
from `t:116` to `e:101`: `116 - 101 = |15| = 15`

From this explanation, we can find `[5, 12, 3, 19. 15]`  that is not increasing. This is why the **update** is not exceeding. If we found any increasing sequence then it'll be exceeding. that's it.

raklaptudirm: Can you give an example of an exceeding word and explain why it is exceeding?

fahimfaisaal: Sure, let's see with the "delete" word, It's exceeding because the **ASCII** distance of each pair is increasing.

Explanation of "**delete**" word:
from `d:117` to `e:101`-> `100- 101= |-1| = 1`
from `e:101`to `l:108`-> `101- 108= |-7| = 7`
from `l:108` to` e:101` -> `108 - 101= |7| = 7`
from `e:101` to `t:116` -> `101 - 116 = |-15| = 15`
from `t:116` to `e:101` -> `116 - 101 = |15| = 15`

From this, we got `[1, 7, 7, 15, 15]` which is increasing. that's why It should call exceeding.

raklaptudirm: I get the misunderstanding we were having now. I will try to resolve it a bit later 👍🏽

fahimfaisaal: @raklaptudirm, Will you resolve this?

raklaptudirm (approved):

PR #987, opened 2022-04-19
Add proper tests for binary search

CarlosZoft:

### Describe your change:

- [x] Add an algorithm?
- [x] Add some test?

### Checklist:


appgurueu (commented):
How does this compare to the current implementation? From what I can see it's merely a less readable iterative version. If it showed better traits than the current one, you may replace it. But I don't see the merit of two largely similar iterative implementations.

raklaptudirm:
@appgurueu since it is being added as a separate implementation, I think we could add it. It is using some more js features, which could be wanted. What do you think?

Search/BinarySearch.js
```diff
@@ -48,8 +48,23 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
   // if low > high => we have searched the whole array without finding the item
   return -1
 }
+function binarySearchIterativeAlternative (arr, x) {
+  let low = -1
+  let high = arr.length
+  while (low < high - 1) {
+    const half = Math.floor((low + high) / 2)
+    // arr[half] === x => low = half else high = half
```
appgurueu, line 56: Superfluous comment that just explains what a ternary does.

Search/BinarySearch.js
```diff
@@ -48,8 +48,23 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
...
+    arr[half] < x ? (low = half) : (high = half)
```
appgurueu, line 57: Should rather be written as `if (arr[half] < x) low = half else high = half`.

Search/BinarySearch.js
```diff
@@ -48,8 +48,23 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
...
+  }
+  // if arr[high] === x => return high else return -1
```
appgurueu, line 59: Again unneeded comment. Please don't explain what the code does, but *why* it does it (something like: "return -1 if the element doesn't match the searched element").

Search/BinarySearch.js
```diff
@@ -89,3 +104,10 @@ export { binarySearchIterative, binarySearchRecursive }
 // binarySearchIterative(stringArr, 'Charlie')
 // binarySearchRecursive(stringArr, 'Zulu')
 // binarySearchIterative(stringArr, 'Sierra')
+// console.log("------------------------------------------")
```
appgurueu, line 107: This code smells like improper testing.

Search/BinarySearch.js
```diff
@@ -48,8 +48,23 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
   // if low > high => we have searched the whole array without finding the item
   return -1
 }
+function binarySearchIterativeAlternative (arr, x) {
+  let low = -1
+  let high = arr.length
+  while (low < high - 1) {
+    const half = Math.floor((low + high) / 2)
```
appgurueu, line 55: This is a minor simplification over what the other implementation does. Consider opening a separate PR, changing the other implementation to `const mid = Math.floor((low + high) / 2)`.

Search/BinarySearch.js
```diff
@@ -48,8 +48,23 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
   // if low > high => we have searched the whole array without finding the item
   return -1
 }
+function binarySearchIterativeAlternative (arr, x) {
+  let low = -1
```
appgurueu, line 52: Consider adding a comment stating that these bounds are both exclusive (which is probably less intuitive than the other impl.).

appgurueu (changes requested):
This PR may indeed be using more JS features (primarily ternaries) but in a wrong manner; they only make the code less readable (except probably for the last ternary). Currently I see only two very tiny advantages over the other impl., the simplification of the term for `mid` and moving the check if found out of the loop; I don't see much of a benefit from the exclusive low & high indices, which is a mere convention. On the other hand however there is rather unreadable code, bad comments and syntactic sugar where it doesn't belong.

The tiny improvements should just be changes to the current implementation.

Search/BinarySearch.js
```diff
@@ -48,8 +48,25 @@ function binarySearchIterative (arr, x, low = 0, high = arr.length - 1) {
...
+}
 
-export { binarySearchIterative, binarySearchRecursive }
+export {
+  binarySearchIterative,
+  binarySearchIterativeAlternative,
+  binarySearchRecursive
+}
 
 /* ---------------------------------- Test ---------------------------------- */
```
raklaptudirm, line 71: Remove these comments and replace them with proper jest tests.

raklaptudirm (changes requested):

appgurueu:
At this point it's similar enough that you should just tidy up the existing implementation (adding proper jest tests) instead.

CarlosZoft:
I did this

appgurueu (changes requested):
From a quick glance the tests look fine to me. Definitely a significant improvement. But please deduplicate the code; you can do something like `funcs = {iterative: binarySearchIterative, recursive: binarySearchRecursive}; for (let name in funcs) runTests(name, funcs[name])` where `runTests` is a parameterized function that runs the tests (which are the same for both implementations).

CarlosZoft:
Nice vision, i will make this!

CarlosZoft:
lgtm now , what you think?

appgurueu (approved):
Thank you for your contribution. LGTM!

raklaptudirm (approved):

PR #1022, opened 2022-05-23
Add the Collatz Conjecture

Exortions:

### Describe your change:

- [x] Add an algorithm?

### Checklist:


Maths/CollatzSequence.js
```diff
@@ -0,0 +1,11 @@
+export function collatz (n) {
+  while (n > 1) {
+    if (n % 2 === 0) {
+      n = n / 2
+    } else {
+      n = 3 * n + 1
+    }
+  }
+
+  return n
```
raklaptudirm, line 10: This program will either never halt or return 1. You should do something more useful with the return value, like counting the number of steps before it reaches the 1, 2, 4 loop.

raklaptudirm (changes requested):
Not everyone is familiar with the collatz conjecture. Add some explanatory comments and Wikipedia links.

Exortions:
Yep, will add those tomorrow.
For now, it's time to sleep 😅

appgurueu (changes requested):
Tests are now broken (and I still don't see cycle detection)

Maths/CollatzSequence.js
```diff
@@ -0,0 +1,30 @@
...
+  const steps = []
+  
+  while (n > 1) {
+    if (n % 2 === 0) {
+      n = n / 2
+    } else {
+      n = 3 * n + 1
+    }
+    
+    steps.push(n)
```
appgurueu, line 26: This excludes the initial `n` - is this intended?

Exortions: Yes, because it tells you the steps from each number, not the original.

Exortions: I can fix it if you would like

Maths/CollatzSequence.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @see [Collatz Conjecture](https://en.wikipedia.org/wiki/Collatz_conjecture)
+ *
+ * @example collatz(1) = { result: 1, steps: [] }
+ * @example collatz(5) = { result: 1, steps: [16, 8, 4, 2, 1] }
+*/
+export function collatz (n) {
+  const steps = []
+
+  while (n > 1) {
```
raklaptudirm, line 19: Instead of `>=`, you should check if `n` is exactly equal to 1, which is more understandable. Also, you should check for all the 3 numbers, 1, 2, and 4, since you want to count the steps till `n` falls in the loop.

raklaptudirm (changes requested):

Maths/CollatzSequence.js
Exortions: No, it should be if n > 1, because the only way the Collatz Conjecture could be disproven is if the number would be less than one.

Exortions: Also if it were `n >= 1`, it would get into an infinite loop, because when you apply the Collatz Conjecture to 1, it will always be 1.

appgurueu:
This is currently unable to disprove the Collatz conjecture as it just won't halt as it lacks cycle detection (which would have to be implemented using a hashmap); it can merely show that certain sequences *don't* disprove the Collatz conjecture.

Exortions:
> This is currently unable to disprove the Collatz conjecture as it just won't halt as it lacks cycle detection (which would have to be implemented using a hashmap); it can merely show that certain sequences _don't_ disprove the Collatz conjecture. I suggest focusing on the sequence rather than the conjecture.

Yeah, so what do you propose I change?

Maths/CollatzSequence.js
raklaptudirm: > No, it should be if n > 1, because the only way the Collatz Conjecture could be disproven is if the number would be less than one.

That is actually wrong, because going to a number less than one following the Collatz rules is impossible. The number only decreases if it is even, and the smallest even number is 2, and 2/2 is 1. So no, the number will never be less than one.

The Collatz conjecture will be disproven if another cycle like the 1, 2, 4 cycle is found, or a number is found such that it keeps on increasing and never falls back down. Computationally, only the first is feasible, as the second is equivalent to the halting problem and is undecidable.

> Also if it were n >= 1, it would get into an infinite loop, because when you apply the Collatz Conjecture to 1, it will always be 1.

We want you to check if it is *not equal to 1*. So when it equals 1 the loop exits.

Exortions: Got it

appgurueu:
Still reads "collatz conjecture" in numerous places...

Exortions:
> Still reads "collatz conjecture" in numerous places...

As in the comments? Or the code?

appgurueu (approved):

raklaptudirm (approved):

PR #1023, opened 2022-05-25
Add pronic number implementation

itsAkshayDubey:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Maths/IsPronic.js
+++ b/Maths/IsPronic.js
@@ -1,0 +1,34 @@
+/*
+ * Author: Akshay Dubey (https://github.com/itsAkshayDubey)
+ * Pronic Number: https://en.wikipedia.org/wiki/Pronic_number
+ * function to check if number is pronic.
+ * return true if number is pronic.
+ * else false
+ */
+
+/**
+ * @function isPronic
+ * @description -> Checking if number is pronic using product of two consecutive numbers
+ * If number is a product of two consecutive numbers, then it is pronic
+ * therefore, the function will return true
+ *
+ * If number is not a product of two consecutive numbers, then it is not pronic
+ * therefore, the function will return false
+ * @param {number} number
+ * @returns {boolean}
+ */
+
+function isPronic(number) {
+
+	for (var i = 0; i <= parseInt(Math.sqrt(number)); i++)
+	
+		// Checking Pronic Number
+		// by multiplying consecutive
+		// numbers
+		if (number == i * (i + 1))
+			return true;
+
+	return false;
+}
+
+export { isPronic }
--- a/Maths/test/IsPronic.test.js
+++ b/Maths/test/IsPronic.test.js
@@ -1,0 +1,13 @@
+import { isPronic } from '../IsPronic'
+
+describe('Testing isPronic function', () => {
+  it('should return if the number is pronic or not', () => {
+    const isPronicNumber = isPronic(2)
+    expect(isPronicNumber).toBe(true)
+  })
+
+  it('should return if the number is pronic or not', () => {
+    const isPronicNumber = isPronic(7)
+    expect(isPronicNumber).toBe(false)
+  })
+})


Maths/IsPronic.js
```diff
@@ -0,0 +1,34 @@
...
+ * therefore, the function will return true
+ *
+ * If number is not a product of two consecutive numbers, then it is not pronic
+ * therefore, the function will return false
+ * @param {number} number
+ * @returns {boolean}
+ */
+
+function isPronic(number) {
+
```
raklaptudirm, line 22: Add a check for whether the number is odd because a pronic number will never be odd.

Maths/IsPronic.js
```diff
@@ -0,0 +1,34 @@
...
+	for (var i = 0; i <= parseInt(Math.sqrt(number)); i++)
```
raklaptudirm, line 23: What is the use of the `parseInt` function here? It is used to convert a string into a number.
```js
parseInt("123") // returns 123
```

Maths/test/IsPronic.test.js
```diff
@@ -0,0 +1,13 @@
+import { isPronic } from '../IsPronic'
+
+describe('Testing isPronic function', () => {
```
raklaptudirm, line 3: Add more test cases.

raklaptudirm (changes requested):
Format your code using `standard.js`. Read the `CONTRIBUTING.md` before contributing.

Maths/IsPronic.js
itsAkshayDubey: Added check of odd number.

Maths/IsPronic.js
itsAkshayDubey: Removed parseInt()

Maths/IsPronic.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {number} number
+ * @returns {boolean}
+ */
+
+function isPronic (number) {
+  if (number % 2 === 1) {
+    return false
+  }
+
+  for (let i = 0; i <= Math.sqrt(number); i++) {
```
raklaptudirm, line 26: Instead of calculating the square root everytime, store it in a variable:
```js
const sqrt = Math.sqrt(number)
```

raklaptudirm (changes requested):

Maths/IsPronic.js
itsAkshayDubey: Added suggested changes to the code.
Thanks.

Maths/test/IsPronic.test.js
```diff
@@ -0,0 +1,18 @@
...
+  it('should return if the number is pronic or not for even number', () => {
+    const isPronicNumber = isPronic(2)
+    expect(isPronicNumber).toBe(true)
+  })
+
+  it('should return if the number is pronic or not for even number', () => {
+    const isPronicNumber = isPronic(4)
```
appgurueu, line 10: Move this to the first `it`.

Maths/IsPronic.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+function isPronic (number) {
+  const sqrt = Math.sqrt(number)
+
+  if (number % 2 === 1) {
+    return false
+  }
+
+  for (let i = 0; i <= sqrt; i++) {
```
appgurueu, line 28: This runs in O(sqrt n) time; it seems unnecessarily inefficient to me. Would checking around the square root not work? `Math.floor(sqrt) * Math.ceil(sqrt) == number` should hold for all pronic numbers.

itsAkshayDubey: Have a look.

appgurueu:
The current implementation is very odd: (1) you don't need to check whether numbers are odd if your equation works (2) your equation is unnecessarily convoluted.

itsAkshayDubey:
> The current implementation is very odd: (1) you don't need to check whether numbers are odd if your equation works (2) your equation is unnecessarily convoluted.

Refactored it, have a look.

appgurueu:
Why use `Math.round` if you can use `Math.floor`/`Math.ceil`? I also failed to consider edge cases properly: `0` is pronic but `1` is not. The following should work:

```javascript
number !== 1 && (number % Math.ceil(Math.sqrt(number)) == 0)
```

itsAkshayDubey:
> Why use `Math.round` if you can use `Math.floor`/`Math.ceil`? I also failed to consider edge cases properly: `0` is pronic but `1` is not. The following should work:
> 
> ```js
> if (number === 0) return true
> const sqrt = Math.sqrt(number)
> return sqrt % 1 !== 0 && Math.ceil(sqrt) * Math.floor(sqrt) === number
> ```
> 
> Please extend the tests to check all example numbers listed on https://oeis.org/A002378 - it's easy to mess up some edge cases. Also check that all other numbers in the range aren't false positives.

Added, have a look.

appgurueu (approved):

raklaptudirm (approved):

PR #1036, opened 2022-06-03
FindSecondLargestElement: Support negative numbers

### LINKED ISSUE #1035: SecondLargestElement Algorithm returns wrong value
for the array [-1, -2, -3, -4] it returns 0

appgurueu:
Trivial change. Fixes #1035.


Sorts/FindSecondLargestElement.js
```diff
@@ -11,7 +11,7 @@
 
 const secondLargestElement = (array) => {
   const largestElement = Math.max(...array)
```
raklaptudirm, line 13: This is a problem of the original algorithm, but shouldn't this be done in the loop? This way of solving it seems like cheating ¯\\\_(ツ)\_/¯

appgurueu: > This is a problem of the original algorithm, but shouldn't this be done in the loop? This way of solving it seems like cheating ¯\_(ツ)_/¯

Nah, it's fine; this is supposed to be idiomatic JS. Otherwise a `findMax` algo should probably be required, or `secondLargestElement` should maintain the two largest elements at a time.

TBF using a heap this may be generalized to findNthLargestElement...

raklaptudirm (approved):

PR #1057, opened 2022-07-14
Wrapping the example matrix with outer array

tmkasun:
### Describe your change:

Without the outer array , given arrays are passed as 4 arguments to the `island` function

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:


appgurueu (approved):
Actual tests would be preferable, but this is definitely an improvement.

raklaptudirm (changes requested):
I think we would prefer to have the comments removed and replaced with actual jest tests.

tmkasun:
Hi @raklaptudirm & @appgurueu ,

I have added a test file for `NumberOfIslands` algorithm, Please review and merge.

raklaptudirm (approved):

tmkasun:
@appgurueu Kind reminder on this

appgurueu (changes requested):
The tests have useless names ("Test Case <n>") currently. Either give them sensible names or move them all into one test block. Please inline to `expect(islands(graph)).toBe(n)`. You might even want to inline `graph` if you find that it doesn't hinder readability.
---

Side note: Why does "number of islands" expect strings; could you change it to use numbers instead?

appgurueu (approved):
(the test case names are pretty redundant though)

tmkasun:
> (the test case names are pretty redundant though)

Yes true, But if I merge them together then it becomes a bit cluttered.

@raklaptudirm Can you please review and merge

raklaptudirm (approved):

PR #1071, opened 2022-07-28
Edit `Perfect{Square, Cube}`

Rudxain:

### Describe your change:

- [ ] Add an algorithm?
- [x] Documentation change?

### Checklist:


appgurueu (dismissed):

Maths/PerfectCube.js
```diff
@@ -2,8 +2,11 @@
  * Author: dephraiim
  * License: GPL-3.0 or later
  *
+ * The integer check acts as a guard clause, and avoids calculating the root if the input is invalid,
+ * it also ensures Infinity is treated as a non-cube.
+ * This uses `round` instead of `floor` or `trunc`, to guard against potential `cbrt` accuracy errors.
  */
 
-const perfectCube = (num) => Math.round(num ** (1 / 3)) ** 3 === num
+const perfectCube = (num) => Number.isInteger(num) && Math.round(Math.cbrt(num)) ** 3 === num
```
raklaptudirm, line 10: You could simplify this to:
<code>
Math.isInteger(Math.cbrt(num))

raklaptudirm (changes requested):

Maths/PerfectCube.js
Rudxain: True. But then that would make the algorithm exclusively work on non-`BigInt` values (it already was, but now the algorithms are completely incompatible). Should I add a `BigInt` variant that uses the "old" algorithm? That would add some educational value, especially because modular arithmetic would be involved (because of `mod 9`). Maybe the `Number` algo would be the "naive" version, while the `BigInt` would act as the optimized variant, both would complement each other

appgurueu: > You could simplify this to: ` Math.isInteger(Math.cbrt(num))`

What if `cbrt` has precision issues (returns something like `2.9999999999` rather than `3`)? Does JS guarantee `Math.cbrt` precision?

Rudxain: > Does JS guarantee `Math.cbrt` precision?

Good question, AFAIK, the accuracy of `Math.cbrt(x)` should be better than `x ** 1/3`, because `1 / 3` is not 100% precise. However, the accuracy is implementation-dependent, so we can't always be 100% sure the answer would be the same on all devices. But (correct me if I'm wrong), this repo is more concerned with the algos rather than implementation details or low-level stuff, so maybe we shouldn't worry about it?

appgurueu: The purpose of this repo is to implement algos correctly and efficiently *in JavaScript*. If the ES spec doesn't guarantee accuracy, it's better to account for inaccuracies.

Rudxain: I think a nice "middle-ground" would be to type it as `Math.round(Math.cbrt(num)) ** 3 === num` without the `isInt` part. I'll reflect the change in `PerfectSquare.js` because this PR is highly "tied" to both files, so I guess it can be an exception to the rule "Only 1 algo per PR"

appgurueu (approved):
Looks fine to me.

raklaptudirm (approved):

PR #1074, opened 2022-08-02
add more functions to DecimalToBinary.js file from Conversions directory

AanthereX:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Conversions/DecimalToBinary.js
+++ b/Conversions/DecimalToBinary.js
@@ -1,13 +1,35 @@
-function decimalToBinary (num) {
-  const bin = []
-  while (num > 0) {
-    bin.unshift(num % 2)
-    num >>= 1 // basically /= 2 without remainder if any
-  }
-  return bin.join('')
+function decimalToBinary(num) {
+    const bin = []
+    while (num > 0) {
+        bin.unshift(num % 2)
+        num >>= 1 // basically /= 2 without remainder if any
+    }
+    return bin.join('')
 }
 
-export { decimalToBinary }
+// using .toString() method
+
+function decimalToBinary_2(dec) {
+    return dec.toString(2)
+}
+
+// using Remainder Quotient method
+
+function decimalToBinary_3(dec) {
+    let bin = 0;
+    let rem, i = 1,
+        step = 1;
+    while (dec != 0) {
+        rem = dec % 2;
+        dec = parseInt(dec / 2);
+        bin = bin + rem * i;
+        i = i * 10;
+    }
+    return bin
+}
+
+
+export { decimalToBinary, decimalToBinary_2, decimalToBinary_3 }
 
 // > decimalToBinary(2)
 // '10'


Conversions/DecimalToBinary.js
```diff
@@ -1,13 +1,35 @@
...
+        num >>= 1 // basically /= 2 without remainder if any
+    }
+    return bin.join('')
 }
 
-export { decimalToBinary }
+// using .toString() method
+
+function decimalToBinary_2(dec) {
+    return dec.toString(2)
```
appgurueu, line 13: This feels like cheating: The purpose of this repo is to demonstrate how this could be implemented without relying on libraries (including the standard library). Perhaps note the existence of a suitable stdlib func in a comment?

Conversions/DecimalToBinary.js
```diff
@@ -16,4 +38,4 @@ export { decimalToBinary }
 // '111'
 
 // > decimalToBinary(35)
-// '100011'
+// '100011'
```
appgurueu, line 41: Please convert to proper tests instead of removing the trailing newline (?)

Conversions/DecimalToBinary.js
```diff
@@ -1,13 +1,35 @@
...
+function decimalToBinary_2(dec) {
+    return dec.toString(2)
+}
+
+// using Remainder Quotient method
+
+function decimalToBinary_3(dec) {
+    let bin = 0;
+    let rem, i = 1,
+        step = 1;
```
appgurueu, line 21: Unused variable.

Conversions/DecimalToBinary.js
```diff
@@ -1,13 +1,35 @@
...
+    while (dec != 0) {
+        rem = dec % 2;
+        dec = parseInt(dec / 2);
```
appgurueu, line 24: Looks like a hacky reimplementation of `decimalToBinary` to me, except instead of using an array to store the digits this uses a base 10 number which may then be stringified in base 10 to produce the desired binary number.

The `parseInt` hack to achieve floor division rather than using `dec >>= 1` is a clear downgrade. You could consider `Math.floor(dec / 2)` for readability, but relying on `parseInt` to perform flooring is hacky IMO.

Conversions/DecimalToBinary.js
```diff
@@ -1,13 +1,35 @@
...
+        bin = bin + rem * i;
```
appgurueu, line 25: You could use `+=` here.

Conversions/DecimalToBinary.js
```diff
@@ -1,13 +1,35 @@
...
+        i = i * 10;
```
appgurueu, line 26: and `*=` here

appgurueu (changes requested):
Overall questionable: Second method just lets JS builtins do the job, third one has poor code quality and is mostly a duplicate of the first one. Leaning towards rejecting this - it probably has to be redone pretty much entirely, if it should be redone at all.

raklaptudirm:
I suggest marking this pr as invalid.

PR #1077, opened 2022-08-02
Fix Affine-Cipher encrypt and drypt func

shidian00:

### Describe your change:
- encrypt func should return alphabet instead of numbers string
- add checking for coprime as if coefficient 'a' is not coprime to 26 will reusultant in invalid result in inverse mod
- modify some test case 

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Ciphers/AffineCipher.js
+++ b/Ciphers/AffineCipher.js
@@ -29,6 +29,27 @@ function inverseMod (a, m) {
 }
 
 /**
+ * GetEuclidGCD Euclidean algorithm to determine the GCD of two numbers
+ * @param {Number} a integer (may be negative)
+ * @param {Number} b integer (may be negative)
+ * @returns {Number} Greatest Common Divisor gcd(a, b)
+ */
+function GetEuclidGCD (a, b) {
+  if (typeof a !== 'number' || typeof b !== 'number') {
+    throw new TypeError('Arguments must be numbers')
+  }
+  if (a === 0 && b === 0) return undefined // infinitely many numbers divide 0
+  a = Math.abs(a)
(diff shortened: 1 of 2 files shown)


Ciphers/AffineCipher.js
```diff
@@ -28,6 +28,27 @@ function inverseMod (a, m) {
   }
 }
 
+/**
+ * GetEuclidGCD Euclidean algorithm to determine the GCD of two numbers
```
appgurueu, line 32: Please don't copy this but rather import the GCD func.

Ciphers/AffineCipher.js
```diff
@@ -44,8 +65,13 @@ function isCorrectFormat (str, a, b) {
     throw new TypeError('Argument str should be String')
   }
 
+  if (!(GetEuclidGCD(a, 26) === 1)) {
+    throw new Error(a + ' is not coprime of 26 !!!')
```
appgurueu, line 69: You don't need the exclamation marks.

shidian00: i replace it with the import coprimecheck

Ciphers/AffineCipher.js
shidian00: Sorry initally when i import the js file, when debug it keep giving me error module not found, but during test it pass i not sure whether it correct or not...

Ciphers/AffineCipher.js
appgurueu: I was referring to the three exclamation marks in the error message.

appgurueu (approved):

PR #1101, opened 2022-09-14
refactor: RotateListRight.js and added tests

10kartik:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:


appgurueu (dismissed):
Looks good. Ideally this should be a method of the `SinglyLinkedList` class though; then you would also be able to leverage the `get` method to obtain an array rather than having to do the tedious `.head`. You could additionally add a constructor that takes a list to shorten your list creation.

raklaptudirm:
@appgurueu Should I await your suggestion being implemented before merging?

appgurueu:
> @appgurueu Should I await your suggestion being implemented before merging?

Not sure. The PR is definitely already an improvement in its current state, but the ideal solution would be what I have suggested.

raklaptudirm:
@10kartik Can you implement the suggested?

10kartik:
Yeah @raklaptudirm, I'll give it a try... But I'm a bit doubtful about the suggested implementation.


> Ideally this should be a method of the `SinglyLinkedList` class though; then you would also be able to leverage the `get` method to obtain an array rather than having to do the tedious `.head`.

As per my understanding @appgurueu, adding a method `RotateListRight` to [LinkedList](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L18l) class will narrow it's usage.


Let's say in a list of length 10, I want to _rotate_ the list after the 4th node, so I'd simply pass the reference of the 5th node as `head` and `k` steps. But if I'll add a method in LinkedList class that only accepts `k` which rotates the list right from the first node, the method will simply lose the essence behind it.

Please correct me if I'm wrong or missing something here.

> You could additionally add a constructor that takes a list to shorten your list creation.

For this, I must add a constructor to the `LinkedList` class, pass an `array/list` as an argument to it and call [addLast](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L45) to create a list, right?

appgurueu:
> As per my understanding @appgurueu, adding a method `RotateListRight` to [LinkedList](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L18l) class will narrow it's usage.
> 
> Let's say in a list of length 10, I want to _rotate_ the list after the 4th node, so I'd simply pass the reference of the 5th node as `head` and `k` steps. But if I'll add a method in LinkedList class that only accepts `k` which rotates the list right from the first node, the method will simply lose the essence behind it.
> 
> Please correct me if I'm wrong or missing something here.

Valid point.

The clean solution to that would be allowing "views" into sublists, comparable to Python or Go slices (although those work on array-based lists). Views would obviously again be linked lists and could thus have `rotateListRight` applied to them.

> > You could additionally add a constructor that takes a list to shorten your list creation.
> 
> For this, I must add a constructor to the `LinkedList` class, pass an `array/list` as an argument to it and call [addLast](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L45) to create a list, right?

Yes.

10kartik:
> The clean solution to that would be allowing "views" into sublists, comparable to Python or Go slices (although those work on array-based lists). Views would obviously again be linked lists and could thus have `rotateListRight` applied to them.

Right! seems reasonable. Okay, so not sure how should I proceed with that [🤔](https://emojipedia.org/thinking-face/). Should I shift the `rotateRightList` function logic to [LinkedList](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L18l) 's method, rotate list by k nodes and reset the head accordingly? Before that, Will add a constructor for creating of LinkedList? Sounds good?

appgurueu:
> > The clean solution to that would be allowing "views" into sublists, comparable to Python or Go slices (although those work on array-based lists). Views would obviously again be linked lists and could thus have `rotateListRight` applied to them.
> 
> Right! seems reasonable. Okay, so not sure how should I proceed with that [thinking](https://emojipedia.org/thinking-face/). Should I shift the `rotateRightList` function logic to [LinkedList](https://github.com/TheAlgorithms/JavaScript/blob/cf0593f4307749ae2c790f500f13f30aa91f23bc/Data-Structures/Linked-List/SinglyLinkedList.js#L18l) 's method, rotate list by k nodes and reset the head accordingly? Before that, I will add a constructor for creating LinkedList... Sounds good?

Yes, sounds good. Methods to create views / sublists can come later (or not at all).

Data-Structures/Linked-List/SinglyLinkedList.js
```diff
@@ -16,9 +16,15 @@ class Node {
 }
 
 class LinkedList {
-  constructor () {
+  constructor (listOfValues) {
     this.headNode = null
     this.length = 0
+
+    if (listOfValues instanceof Array) {
+      for (let i = 0; i < listOfValues.length; i++) {
```
appgurueu, line 24: Could use `for` - `of` here

appgurueu (approved):

Data-Structures/Linked-List/SinglyLinkedList.js
raklaptudirm: Should I was for this improvement?

appgurueu (approved):

raklaptudirm (approved):

PR #1102, opened 2022-09-17
algorithm: Iterative (and in-place) BFS for binary trees

kho-kho-kho:
Wrote a similar version of this to help visualize some of the testing I had done recently with the AVLTree.

It's helpful for folks to know that recursion isn't the end-all-be-all solution for tree-based problems. Recursive solutions are great for academia, but depending on the type of commercial development they'll be doing recursion may not always be an acceptable option.

### Describe your change:

- [x] Add an algorithm?
- [x] Documentation change?

### Checklist:

### DIFF
--- a/Trees/BreadthFirstTreeTraversal.js
+++ b/Trees/BreadthFirstTreeTraversal.js
@@ -17,7 +17,27 @@ class BinaryTree {
     this.traversal = []
   }
 
-  breadthFirst () {
+  breadthFirstIterative () {
+    this.traversal = []
+    if (this.root) {
+      this.traversal.push(this.root)
+    }
+    let currentSize = this.traversal.length
+    for (let i = 0; i < currentSize; i++) {
+      if (this.traversal[i].left) {
+        this.traversal.push(this.traversal[i].left)
+      }
+      if (this.traversal[i].right) {
+        this.traversal.push(this.traversal[i].right)
+      }
+      this.traversal[i] = this.traversal[i].data
+      currentSize = this.traversal.length
+    }
+    return this.traversal
+  }
+
+  breadthFirstRecursive () {
+    this.traversal = []
     const h = this.getHeight(this.root)
     for (let i = 0; i !== h; i++) {
       this.traverseLevel(this.root, i)
--- a/Trees/BreadthFirstTreeTraversal.js
+++ b/Trees/BreadthFirstTreeTraversal.js
@@ -22,16 +22,15 @@ class BinaryTree {
     if (this.root) {
       this.traversal.push(this.root)
     }
-    let currentSize = this.traversal.length
-    for (let i = 0; i < currentSize; i++) {
-      if (this.traversal[i].left) {
-        this.traversal.push(this.traversal[i].left)
+    for (let i = 0; i < this.traversal.length; i++) {
+      const currentNode = this.traversal[i]
+      if (currentNode.left) {
+        this.traversal.push(currentNode.left)
       }
-      if (this.traversal[i].right) {
-        this.traversal.push(this.traversal[i].right)
+      if (currentNode.right) {
+        this.traversal.push(currentNode.right)
       }
-      this.traversal[i] = this.traversal[i].data
-      currentSize = this.traversal.length
+      this.traversal[i] = currentNode.data
     }
     return this.traversal
   }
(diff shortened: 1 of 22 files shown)


Trees/BreadthFirstTreeTraversal.js
```diff
@@ -17,7 +17,27 @@ class BinaryTree {
     this.traversal = []
   }
 
-  breadthFirst () {
+  breadthFirstIterative () {
+    this.traversal = []
```
appgurueu, line 21: I definitely don't like storing the traversal in an instance variable; this should be a local variable and returned (applies for the other traversal methods as well).

Trees/BreadthFirstTreeTraversal.js
```diff
@@ -17,7 +17,27 @@ class BinaryTree {
...
+    if (this.root) {
+      this.traversal.push(this.root)
+    }
+    let currentSize = this.traversal.length
```
appgurueu, line 25: Any reason why this needs to be a variable?

Trees/BreadthFirstTreeTraversal.js
```diff
@@ -17,7 +17,27 @@ class BinaryTree {
...
+    for (let i = 0; i < currentSize; i++) {
+      if (this.traversal[i].left) {
```
appgurueu, line 27: Could shorten the body by doing `let currentNode = traversal[i]`

appgurueu (commented):
The algorithm is fine but inherits some poor code quality.

Trees/BreadthFirstTreeTraversal.js
kho-kho-kho: I agree.

The only reason I can see for doing this was that the original dev was implementing specifically a BFS traversal, so maybe they thought it would be meaningful to store the order. IMO it makes more sense to "do something" at each node, but that's less relevant WRT academia.

Trees/BreadthFirstTreeTraversal.js
kho-kho-kho: You mean "currentSize"?

I wanted to manually update the variable myself so that the for-loop would 100% always check the value I expected. You wouldn't normally update the conditional dynamically in a for-loop (besides the index variable 'i'), so this was just to ensure I got the behavior I wanted without relying on some less obvious implementation detail regarding javascript.

Trees/BreadthFirstTreeTraversal.js
kho-kho-kho: Haha .. That's actually how I had it before making my final changes.

I ultimately decided I'd rather save 1 line of code while sacrificing some width of each line. My motivation was keeping the algo as short as possible since it wasn't going to be that complicated to begin with.

Not a huge deal though .. I can change it back.

Trees/BreadthFirstTreeTraversal.js
kho-kho-kho: I was in the process of getting rid of the instance var, but then realized we'd have to alter the original algorithm a bit as well. You'd need to pass in some local variable that gets modified correctly throughout the recursion.

I'm fine to look into this, but can we get the iterative implementation committed first? I can make the other changes we've discussed, but WRT to changing the original algo I'd rather that happen in a separate commit.

appgurueu (approved):
Looks good, can be iterated upon as you said.

Trees/BreadthFirstTreeTraversal.js
appgurueu: Hmm, I don't think list length is "a less obvious implementation detail"...

raklaptudirm:
@appgurueu All the conversations are resolved I suppose?

Trees/BreadthFirstTreeTraversal.js
appgurueu: Sure, we can delay that for now.

appgurueu:
> @appgurueu All the conversations are resolved I suppose?

Yes. It's minor enough to be postponed.

kho-kho-kho:
> > @appgurueu All the conversations are resolved I suppose?
> 
> Yes. It's minor enough to be postponed.

@appgurueu Appreciate the review!

@raklaptudirm Good to merge now?

raklaptudirm (approved):

PR #1116, opened 2022-10-02
Added an alternative method for swapping elements in selection sort

ayeola05:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Sorts/SelectionSort.js
+++ b/Sorts/SelectionSort.js
@@ -20,16 +20,38 @@ export const selectionSort = (list) => {
     }
     // Number of passes
     let min = i // min holds the current minimum number position for each pass; i holds the Initial min number
-    for (let j = i + 1; j < length; j++) { // Note that j = i + 1 as we only need to go through unsorted array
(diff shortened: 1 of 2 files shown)


appgurueu (changes requested):
This is really just changing swapping to use a temporary variable rather than a temporary array and array destructuring. You could consider this as a performance-enhancing change for naive JS engines. Duplicating the algo is not worth it. Edit the initial algo instead.

appgurueu (approved):

appgurueu:
There's a tradeoff to be made here between "code quality" and performance. Swapping using a temporary variable also is an idiom common to various languages that may be worth showing. Then again, doing a swap in a single line is much more compact & readable in terms of vertical space - at the cost of creating a temporary array, which V8 optimizes out however (IIRC SpiderMonkey did not do this, but probably does it by now).

All in all I guess I have to to agree with @raklaptudirm. In the end it won't matter on V8 (Node.js) anyways, so you could even use the code in its current state in production as long as you don't expect it to support IE ;)

Issue #1121, opened 2022-10-05
DateToDay.js giving wrong day


chrisjwaddell:
DateToDay("1/1/2020") returns "Friday" but its "Wednesday". Other basic tests show the wrong day as well.

alpesh-rama:
Please assign me to this issue. Thank you

rajprogrammerbd:
I want to work on please, Thank you.

anshul137:
I would like to solve this problem.
thanks

raji2004:
please assign it to me

chrisgalvan:
I have also submitted a solution to current bug

Hanee-Solanki:
Hi, this is my first time contributing to an open-source project. I wanted to know if we have to use the same algorithm to calculate the day or if we could use a different one. Please advise.

raklaptudirm:
There are already several pull requests fixing this issue. I will advise people to not take this issue up any more.

Gauri-AB:
Hii Sir, pls assign this issue to me.

appgurueu:
> Hii Sir, pls assign this issue to me.

We already have two working fixes for this issue.

PR #1122, opened 2022-10-05
fix function of DateToDay and the tests

rajprogrammerbd:
This pull request solves the issue of #1121


sumantpaldm:
my team has solved this problem without using the date feature
please consider our solution too.

sumantpaldm:
we have used totally different approach to fix this issue

PR #1125, opened 2022-10-05
fix: DateToDay

Chetan07j:

### Describe your change:

- [X] Fix a bug or typo in an existing algorithm?

### Checklist:

- [X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Javascript/blob/master/CONTRIBUTING.md).
- [X] This pull request is all my own work -- I have not plagiarized.
- [X] I know that pull requests will not be merged if they fail the automated tests.
- [X] This PR only changes one algorithm file. To ease review, please open separate PRs for separate algorithms.
- [X] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

### DIFF
--- a/Conversions/DateToDay.js
+++ b/Conversions/DateToDay.js
@@ -41,7 +41,7 @@ const DateToDay = (date) => {
   const century = Math.floor(year / 100)
 
   // Apply the algorithm shown above
-  const weekDay = (day + Math.floor((month + 1) * 2.6) + yearDigit + Math.floor(yearDigit / 4) + Math.floor(century / 4) - (2 * century)) % 7
+  const weekDay = (day + Math.floor((month + 1) * 2.6) + yearDigit + Math.floor(yearDigit / 4) + Math.floor(century / 4) + (5 * century)) % 7
 
   // There is possibility that weekDay is in decimal, so Math.floor() needed
   // return the weekDay name.
--- a/Conversions/DateToDay.js
+++ b/Conversions/DateToDay.js
@@ -12,32 +12,8 @@
     Algorithm & Explanation : https://en.wikipedia.org/wiki/Zeller%27s_congruence
 */
 
-// March is taken as the first month of the year.
-const calcMonthList = {
-  1: 13,
-  2: 14,
-  3: 3,
-  4: 4,
-  5: 5,
-  6: 6,
-  7: 7,
-  8: 8,
-  9: 9,
-  10: 10,
-  11: 11,
-  12: 12
-}
-
-// show the week day in a number : Sunday - Saturday => 0 - 6
-const daysNameList = { // weeks-day
-  1: 'Sunday',
-  2: 'Monday',
-  3: 'Tuesday',
-  4: 'Wednesday',
-  5: 'Thursday',
-  6: 'Friday',
-  0: 'Saturday'
-}
+// Array holding name of the day: Saturday - Sunday - Friday => 0 - 1 - 6
+const daysNameArr = ['Saturday', 'Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday']
 
 const DateToDay = (date) => {
   // firstly, check that input is a string or not.
@@ -51,19 +27,25 @@ const DateToDay = (date) => {
     return new TypeError('Date is not valid.')
   }
 
-  // In case of Jan and Feb, we consider it as previous year
+  // In case of Jan and Feb:
+  // Year: we consider it as previous year
   // e.g., 1/1/1987 here year is 1986 (-1)
+  // Month: we consider value as 13 & 14 respectively
   if (month < 3) {
-    year -= 1
+    year--
+    month += 12
   }
 
   // divide year to century and yearDigit value.
   const yearDigit = (year % 100)
   const century = Math.floor(year / 100)
+
   // Apply the algorithm shown above
-  const weekDay = Math.abs((day + Math.floor((calcMonthList[month] + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
+  const weekDay = Math.abs((day + Math.floor((month + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
+
+  // There is possibility that weekDay is in decimal, so Math.floor() needed
   // return the weekDay name.
-  return daysNameList[Math.floor(weekDay)]
+  return daysNameArr[Math.floor(weekDay)]
 }
 
 // Example : DateToDay("18/12/2020") => Friday
--- a/Conversions/DateToDay.js
+++ b/Conversions/DateToDay.js
@@ -40,10 +40,21 @@ const DateToDay = (date) => {
   const yearDigit = (year % 100)
   const century = Math.floor(year / 100)
 
-  // Apply the algorithm shown above
+  /**
+   * Algorithm implementation
+   *
+   * In the mathematics modulo operations, truncated division is used mostly in most of programming languages.
+   * Truncation defines quotient part (integer part) q = trunc(a/n), remainder has same sign as dividend.
+   * -2 mod 7 return result of -2 => console.log(-2 % 7) => -2
+   *
+   * To overcome this problem, to ensure positive numerator, formula is modified by replacing -2J with +5J (J => century)
+   *
+   * Following example shows issue with modulo division
+   * 1. For date 2/3/2014 with old formula - (2 * century) weekDay comes as -6 and wrong day
+   * 2. With computer logic + (5 * century) it gives proper valid day as Sunday
+   */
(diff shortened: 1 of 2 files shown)


Chetan07j:
@appgurueu Could you please share what are your thoughts on this PR?
It will be really helpful for me if there are any improvements I can take care of next time.
Thanks

Conversions/DateToDay.js
```diff
@@ -9,34 +9,34 @@
     algorithm shown below gives us the number of the day and
     finally converts it to the name of the day.
 
-    Algorithm & Explanation : https://cs.uwaterloo.ca/~alopez-o/math-faq/node73.html
+    Algorithm & Explanation : https://en.wikipedia.org/wiki/Zeller%27s_congruence
 */
 
 // March is taken as the first month of the year.
 const calcMonthList = {
```
appgurueu, line 16: This mapping should (1) use an array and (2) definitely needs a comment; "March is taken as first month of the year" explains nothing. I prefer the PRs that get rid of it entirely.

Conversions/DateToDay.js
```diff
@@ -9,34 +9,34 @@
...
+  6: 6,
+  7: 7,
+  8: 8,
+  9: 9,
+  10: 10,
+  11: 11,
+  12: 12
 }
 
 // show the week day in a number : Sunday - Saturday => 0 - 6
```
appgurueu, line 31: This comment is now wrong.

Conversions/DateToDay.js
```diff
@@ -45,18 +45,25 @@ const DateToDay = (date) => {
...
   // check the data are valid or not.
-  if (day < 0 || day > 31 || month > 12 || month < 0) {
+  if (day < 1 || day > 31 || month > 12 || month < 1) {
     return new TypeError('Date is not valid.')
   }
+
+  // In case of Jan and Feb, we consider it as previous year
+  // e.g., 1/1/1987 here year is 1986 (-1)
+  if (month < 3) {
+    year -= 1
```
appgurueu, line 57: `year--`

Conversions/DateToDay.js
```diff
@@ -9,34 +9,34 @@
...
+  7: 7,
+  8: 8,
+  9: 9,
+  10: 10,
+  11: 11,
+  12: 12
 }
 
 // show the week day in a number : Sunday - Saturday => 0 - 6
 const daysNameList = { // weeks-day
```
appgurueu, line 32: This should be an array.

Conversions/DateToDay.js
Chetan07j: Noted.
Updated code as per comment.

I had thought of making it array as you suggested, like below:

```js
/**
 * This array holds months number
 * Considering March as the first month of the year
 * For January & February we store value as 13 & 14 respectively,
 * we consider these two from last year
 * And for other months we store exact month number
 */
const monthArray = [13, 14, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
```

/# And then use it as **monthArray[month - 1]**
```js
const weekDay = Math.abs((day + Math.floor((monthArray[month - 1] + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
```

Then updated the code and remove this dependency completely.

Conversions/DateToDay.js
Chetan07j: Noted and updated comment to more meaningful

Conversions/DateToDay.js
Chetan07j: Done.

Conversions/DateToDay.js
Chetan07j: Done. Made it as an array, very good suggestion.

Conversions/DateToDay.js
```diff
@@ -51,19 +27,25 @@ const DateToDay = (date) => {
...
+    month += 12
   }
 
   // divide year to century and yearDigit value.
   const yearDigit = (year % 100)
   const century = Math.floor(year / 100)
+
   // Apply the algorithm shown above
-  const weekDay = Math.abs((day + Math.floor((calcMonthList[month] + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
+  const weekDay = Math.abs((day + Math.floor((month + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
```
appgurueu, line 44: Which formula are you implementing here?

Conversions/DateToDay.js
```diff
@@ -51,19 +27,25 @@ const DateToDay = (date) => {
...
+
+  // There is possibility that weekDay is in decimal, so Math.floor() needed
   // return the weekDay name.
-  return daysNameList[Math.floor(weekDay)]
+  return daysNameArr[Math.floor(weekDay)]
```
appgurueu, line 48: None of the formulas I've seen is `floor(abs(...))`.

Conversions/DateToDay.js
Chetan07j: https://en.wikipedia.org/wiki/Zeller%27s_congruence

appgurueu: You didn't answer my question. Which formula - there are multiple formulas for different calendars on that page. You will want the one which (1) doesn't have problems with JS mod and (2) is for the Gregorian (not the Julian) calendar:
Formula

Chetan07j: Sorry, misunderstood it.
I have implemented this one:
image

appgurueu: No, you did not:

```javascript
Math.floor((day + Math.floor((month + 1) * 2.6) + yearDigit + (yearDigit / 4) + (century / 4) - (2 * century)) % 7)
```

* Why is there a `floor` around it? There is no floor around the formula.
* There must be a `floor` around `K/4` and `J/4`; you **may not** apply this afterwards
* This won't work for all cases due to JavaScript's [modulo operator](https://en.wikipedia.org/wiki/Zeller's_congruence#Basic_modification)

Chetan07j: Updated, could you please check

appgurueu: You have addressed the first two points, but how have you addressed the last point?

Chetan07j: I have now addressed point 3 as well, as per explanation given [here](https://en.wikipedia.org/wiki/Zeller's_congruence#Basic_modification)

> The formulas rely on the mathematician's definition of [modulo](https://en.wikipedia.org/wiki/Modulo_operation) division, which means that −2 mod 7 is equal to positive 5. Unfortunately, in the truncating way most computer languages implement the remainder function, −2 mod 7 returns a result of −2. So, to implement Zeller's congruence on a computer, the formulas should be altered slightly to ensure a positive numerator. The simplest way to do this is to replace − 2J by + 5J

image

Conversions/DateToDay.js
```diff
@@ -9,54 +9,43 @@
...
+  if (month < 3) {
+    year--
+    month += 12
+  }
+
   // divide year to century and yearDigit value.
   const yearDigit = (year % 100)
   const century = Math.floor(year / 100)
+
   // Apply the algorithm shown above
```
appgurueu, line 43: Please add a note explaining that Java's modulo operator requires an adaption of the formula.

Conversions/DateToDay.js
```diff
@@ -9,54 +9,43 @@
...
-  const weekDay = Math.abs((day + Math.floor((2.6 * calcMonthList[month]) - 0.2) - (2 * century) + yearDigit + Math.floor(yearDigit / 4) + Math.floor(century / 4)) % 7)
+  const weekDay = (day + Math.floor((month + 1) * 2.6) + yearDigit + Math.floor(yearDigit / 4) + Math.floor(century / 4) + (5 * century)) % 7
+
+  // There is possibility that weekDay is in decimal, so Math.floor() needed
```
appgurueu, line 46: This comment is now outdated.

Conversions/DateToDay.js
Chetan07j: Done. Updated comment

Conversions/DateToDay.js
Chetan07j: Done.

Conversions/DateToDay.js
```diff
@@ -40,10 +40,21 @@ const DateToDay = (date) => {
   const yearDigit = (year % 100)
   const century = Math.floor(year / 100)
 
-  // Apply the algorithm shown above
+  /**
```
appgurueu, line 43: Don't copy verbatim from Wikipedia.

Chetan07j: I have tried to explain the modulo division, please check and let me know if it makes sense.
Also, I have updated the commit message as `comment update` instead doc (_which was incorrect in this case_).

appgurueu (approved):

raklaptudirm (approved):

Chetan07j:
Thank you @appgurueu @raklaptudirm for merging my PR.

Thank you @appgurueu for all your valuable review comments and suggestion, I really learned a lot from you.

PR #1142, opened 2022-10-07
solution: ProjectEuler-007

pomkarnath98:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Project-Euler/Problem007.js
+++ b/Project-Euler/Problem007.js
@@ -16,12 +16,12 @@ export const nthPrime = (n) => {
   }
 
   let count = 0
-  let candidateValue = 2
+  let candidateValue = 1
   while (count < n) {
+    candidateValue++
     if (PrimeCheck(candidateValue)) {
       count++
     }
-    candidateValue++
   }
-  return candidateValue - 1
+  return candidateValue
 }
--- a/Project-Euler/Problem007.js
+++ b/Project-Euler/Problem007.js
@@ -1,0 +1,27 @@
+import { PrimeCheck } from '../Maths/PrimeCheck.js'
+
+/**
+ * Find nth Prime Number
+ *
+ * P.S.(Project Euler - 007):
+ * By listing the first six prime numbers: 2, 3, 5, 7, 11, and 13, we can see that the 6th prime is 13.
+ * What is the 10001st prime number?
+ *
+ * @param {Number} n
+ * @returns {Number} returns the nth prime number
+ */
+export const nthPrime = (n) => {
+  if (n < 1) {
+    return 'Invalid Input'
+  }
+
+  let count = 0
+  let inc = 2
+  while (count < n) {
+    if (PrimeCheck(inc)) {
+      count++
+    }
+    inc++
+  }
+  return inc - 1
+}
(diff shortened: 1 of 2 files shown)


Project-Euler/Problem007.js
```diff
@@ -0,0 +1,27 @@
...
+ * P.S.(Project Euler - 007):
+ * By listing the first six prime numbers: 2, 3, 5, 7, 11, and 13, we can see that the 6th prime is 13.
+ * What is the 10001st prime number?
+ *
+ * @param {Number} n
+ * @returns {Number} returns the nth prime number
+ */
+export const nthPrime = (n) => {
+  if (n < 1) {
+    return 'Invalid Input'
```
appgurueu, line 15: Should be an error.

Project-Euler/Problem007.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  let count = 0
+  let inc = 2
+  while (count < n) {
```
appgurueu, line 20: Might as well decrement `n` here (`while (n--)` should work)

Project-Euler/Problem007.js
```diff
@@ -0,0 +1,27 @@
...
+    if (PrimeCheck(inc)) {
+      count++
+    }
+    inc++
+  }
+  return inc - 1
```
appgurueu, line 26: If you increment at the beginning of the loop, you don't have to decrement here

Project-Euler/Problem007.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Number} n
+ * @returns {Number} returns the nth prime number
+ */
+export const nthPrime = (n) => {
+  if (n < 1) {
+    return 'Invalid Input'
+  }
+
+  let count = 0
+  let inc = 2
```
appgurueu, line 19: `inc`? Please name this something like "candidate" or the like

Project-Euler/Problem007.js
pomkarnath98: @appgurueu I am confused here. We're incrementing the value of `count` only if the number is prime. So how can we do that if we will do `while (n--)`?

Project-Euler/Problem007.js
pomkarnath98: Sorry @appgurueu  but again I got confused. Incrementing at the beginning will also do the same at the end of the loop. In both cases, we have to decrease the value.

Project-Euler/Problem007.js
appgurueu: Nevermind, you're right.

Project-Euler/Problem007.js
appgurueu: No. If you increment at the beginning of the loop (and start with `candidate = 1`), when `PrimeCheck(candidate)` being truthy triggers `count++`, you don't enter the next iteration and thus `candidate` holds the correct number rather than the last prime number plus one.

pomkarnath98: My bad! Got your point. I will make the changes.
Definitely, there're many ways to solve the it.

appgurueu (approved):

raklaptudirm (approved):

PR #1150, opened 2022-10-08
feat: added Project Euler Problem 13

byt3h3ad:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:


appgurueu:
Your checks are failing.

byt3h3ad:
all errors have been fixed

byt3h3ad:
@appgurueu @raklaptudirm any changes needed?

appgurueu:
I don't think this algorithm is correct in the general case. The numbers given are well below what JavaScript numbers (64-bit floats) can safely represent; that is, they exceed `MAX_SAFE_INTEGER` by a lot and get truncated to the first `n` of their most significant digits. When you now add all these numbers, you have some error. Can you rule out the possibility that this error might cause the results to be wrong? For the given test case it might work, but does it work in general? Even if it did, it wouldn't be apparent why (and would be deeply reliant on floating-point intricacies).

The proper solution is to use big integers to exactly represent the given numbers.

byt3h3ad:
@appgurueu thanks for the feedback! I have now used BigInt, and turns out there was indeed an error xD.

Project-Euler/Problem013.js
```diff
@@ -14,9 +14,9 @@
 
 export const largeSum = (nums) => {
   const arr = nums.split('\n') // convert to array of strings
-  let sum = 0
+  let sum = 0n
   for (const item of arr) {
-    sum += parseInt(item)
+    sum += BigInt(item)
   } // calculate the sum
   const res = parseInt(sum.toString().substring(0, 11).replace('.', '')) // convert the sum to a string, get the first 10 digits and remove the decimal
```
appgurueu: You don't need to remove the decimal anymore

appgurueu (approved):

Project-Euler/Problem013.js
byt3h3ad: thanks!

raklaptudirm (approved):

Issue #1193, opened 2022-10-15
Project-Euler/test/Problem044.test.js runs for ~ 17 minutes


defaude:
I just stumbled upon this gem while executing the tests locally and then wondering what's going on.

See https://github.com/TheAlgorithms/JavaScript/actions/runs/3255189818/jobs/5344248296#step:5:161

@raklaptudirm what do you think? Should we
1. disable the test so it's skipped and doesn't clog up our CI
2. figure out **why** this immense delay happens, fix the test (or the implementation) and re-enable the test once we're good

defaude:
It's this one:

```js
  // Project Euler Second Value for Condition Check
  test('if the number is greater or equal to 2167', () => {
    expect(problem44(2167)).toBe(8476206790)
  })
```

I **guess** it's because this is running into really, really high numbers, maybe? Neither the CPU nor the RAM usage goes up substantially (at least on my machine) but for some reason, this takes **ages** to complete.

defaude:
I just realized this is even more severe as assumed initially: Since we're running the test in a pre-commit hook, the simple act of **just committing changes locally** becomes a minute-long ordeal...

appgurueu:
> I just realized this is even more severe as assumed initially: Since we're running the test in a pre-commit hook, the simple act of **just committing changes locally** becomes a minute-long ordeal...

OOF. Ideally CI & hooks should only run tests on the files that were changed.

raklaptudirm:
Jest usually caches tests so the precommit hook should not be an issue.

utkarsh-shrivastav77:
Is this issue still open to contribute

appgurueu:
> export { problem44 } instead of this can I use console.log for the answer

No, that doesn't improve the performance at all and just harms code quality.

defaude:
While we changed the process so that the pipeline for pull requests only executes changed test files, the issue with the super-long-running test execution. This hinders not only the "regular" pipeline regarding runtime: It also kind of deters people from just running the tests locally as it takes forever. As a result, we have three failing tests right now :)

=> I took the liberty to just `.skip` the long-running test. This way, we still **have** the test case for those who want to play around with it and for documentation's sake. But we don't **run** it every single time.

PR #1194, opened 2022-10-15
tests: only test changed packages

defaude:
Relates to #1193


appgurueu (changes requested):
This is akin to outcommenting the test.

Project Euler solutions can be expected to take some time (should be < one minute though, ideally the algo should be optimized).

If you do not want to run this test, simply exclude the directory when running tests using Jest.

defaude:
However, having this long-running tests clogs up everyone's workflow and not "just" the CI, therefore I recommend to **not** just uncommenting the test but rather explicitly skipping it. This way, it's still visible in the console. Finding a solution to the runtime behavior is step 2 - but meanwhile, everyone else can continue working without that delay.

What do you think?

appgurueu (dismissed):

appgurueu:
We'll have to leave the issue open to find a proper fix eventually however.

defaude:
Yup! 👍  I mereley mentioned the issue no. in the commit message but didn't _close_ it.

raklaptudirm:
Ideally, we would want to run the tests in the changed directories only.

defaude:
Jest has an `onlyChanged` option that I've set up in the pre-commit hook. If that's what you like, we could also switch to that behavior for the CI, as well.

raklaptudirm:
That could be great!

defaude:
@raklaptudirm I've updated the CI config, as well 👍

CONTRIBUTING.md
```diff
@@ -107,6 +108,12 @@ You can also start Jest in "watch" mode:
 npm test -- --watchAll
 ```
 
+We also prepared a helper script that runs tests only for changed files:
+
+```shell
+npm run test-watch
```
appgurueu, line 114: Shouldn't this be `npm run test-changed`?

defaude: Of course (the day was too long, sorry... 😅) I'll update my PR immediately.

appgurueu (dismissed):
Thank you for your contribution!

raklaptudirm:
The pr looks fine, but I had an idea. I think we would want to run all the tests in the master branch, but run tests on only the changed files in the prs. What do you all think?

defaude:
I've updated the Workflow accordingly - please take a look 😀

raklaptudirm:
> I guess I could set that up with GitHub Actions, yeah. But as long as #1193 is not solved, the `master` build will take forever to complete.

The master build should not be an issue as no one will need to wait for that to complete.

.github/workflows/Ci.yml
```diff
@@ -20,8 +21,13 @@ jobs:
...
 
-      - name: 🧪 Run tests
-        run: npm test
+      - name: 🧪 Run all tests
+        if: ${{ github.event_name == 'push' }}
+        run: npm run test
+
+      - name: 🧪 Run tests for changed files only
+        if: ${{ github.event_name == 'pull_request' }}
+        run: npm run test
```
raklaptudirm, line 30: This will be test-changed i guess

defaude: You're absolutely right....... 😨

appgurueu (approved):

PR #1196, opened 2022-10-15
first knight condicional

EduardoLariosEspinoza:
Modified a condition that return form itself a boolean

### DIFF
--- a/Backtracking/KnightTour.js
+++ b/Backtracking/KnightTour.js
@@ -1,12 +1,12 @@
 // Wikipedia: https://en.wikipedia.org/wiki/Knight%27s_tour
 
 class OpenKnightTour {
-  constructor (size) {
+  constructor(size) {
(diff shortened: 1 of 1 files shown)


Backtracking/KnightTour.js
```diff
@@ -22,37 +22,37 @@ class OpenKnightTour {
...
     return !this.board.map(row => row.includes(0)).includes(true)
   }
 
-  solve () {
+  solve() {
     // function to find the solution for the given board
     for (let i = 0; i < this.size; i++) {
       for (let j = 0; j < this.size; j++) {
-        if (this.solveHelper([i, j], 0)) return true
+        return this.solveHelper([i, j], 0)
```
appgurueu, line 34: This is wrong. The previous code would only return early if this was truthy. Now it will return early as soon as this returns, even if it's falsy.

PR #1221, opened 2022-10-21
Added algorithm CheckHeterogram

KevinLindh:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/String/CheckHeterogram.js
+++ b/String/CheckHeterogram.js
@@ -1,0 +1,24 @@
+// A [Heterogram](https://en.wikipedia.org/wiki/Heterogram_(literature)) is a word, phrase or sentence in which no letter of the alphabet occurs more than once.
+
+/**
+ * @function checkHeterogram
+ * @param {string} str
+ * @returns {boolean}
+ * @description - check if a string is a heterogram with the help of Regex
+ * @example - checkHeterogram('hero') => true
+ * @example - checkHeterogram('CoMpUtEr') => true
+ * @example - checkHeterogram('Great work!!') => true
+ * @example - checkHeterogram('Aba') => false
+ */
+
+const checkHeterogram = (str) => {
+  // Check that input is a string
+  if (typeof str !== 'string') {
+    throw new TypeError('Argument not a string.')
+  }
+
+  // finally check that there are no repeated letters (case insensitive) and return the answer.
+  return !str.match(/([a-z]).*\1/i)
+}
+
+export { checkHeterogram }
--- a/String/test/CheckHeterogram.test.js
+++ b/String/test/CheckHeterogram.test.js
@@ -1,0 +1,38 @@
+import { checkHeterogram } from '../CheckHeterogram'
+
+describe('Testing checkHeterogram', () => {
+  it('expects to throw type error if given a non string argument', () => {
+    expect(() => checkHeterogram(0)).toThrow()
+  })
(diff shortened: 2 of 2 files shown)


String/test/CheckHeterogram.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe('Testing checkHeterogram', () => {
+  it('expects to throw type error if given a non string argument', () => {
+    expect(() => checkHeterogram(0)).toThrow()
+  })
+  it('expects to throw type error if given a non string argument', () => {
+    expect(() => checkHeterogram(true)).toThrow()
+  })
+  it('expects to return true if the argument is a heterogram', () => {
+    const SUT = checkHeterogram('hero')
```
appgurueu, line 11: please get rid of the `SUT` variable in all tests - it adds no value

String/CheckHeterogram.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+const checkHeterogram = (str) => {
+  // Check that input is a string
+  if (typeof str !== 'string') {
+    throw new TypeError('Argument not a string.')
+  }
+
+  // finally check that there are no repeated letters (case insensitive) and return the answer.
+  return !str.match(/([a-z]).*\1/i)
```
appgurueu, line 21: heh, indeed, the JS regex engine is powerful enough for this; the reason for this is that JS RegExes actually aren't regular - while you could build a regular expression for this, this uses something more powerful (think pushdown automaton / CFG)

I would advise against RegEx for two reasons:
1. It hides all the relevant implementation details (basically the core of the algorithm; I'd use a simple hash map of letters here and loop over the characters of the string and inserting their lowercase equivalents until you hit a duplicate at which point you return false; if it doesn't happen you have true as a default return).
2. As a consequence of 1, you often get subpar time (& sometimes space) complexity. What is the time complexity of this implementation? I can't tell (and neither can you) without knowing the intricacies of the regex engine used. In this case it is quite likely that the worst case is exponential since (unless the JS regex engine is very smart) this can't be turned into a DFA to be applied to the input in linear time. On the other hand, a solution that manually does the work is guaranteed linear time.

String/test/CheckHeterogram.test.js
KevinLindh: Thank you for noticing this detail. It was my first time implementing testing, so clearly made a quite foolish error looking back at it now.

String/CheckHeterogram.js
KevinLindh: Thank you for taking the time to go through this thoroughly. I've heard of regex quite often being referred to as a black box because of its mystic nature and it being quite difficult to tell exactly what goes on "under the hood". I checked some other solutions in the repo which used a lot of regexes and I guess my inner code golfer got the better of me. 

Since this repo is meant to increase the understanding of the time and space complexities of different solutions, the character length of the solutions probably shouldn't be the main focus. Breaking the solutions down into as many small and manageable components as possible while still reducing the amount of unnecessary code should of course be the main focus. From my own experience, regex is really interesting but on multiple occasions, I have had to decode some of my old regexes because they are just too hard to understand sometimes.

I tried my best to implement a solution that is easier to understand and walk through logically (good old for loop). I'm still learning about and implementing "optimal" algorithms so hopefully, this would be considered a better solution.

PR #1264, opened 2022-11-24
Update removeAt function

TedMwai:
Previous removeAt function would throw a type error if the index is equals to the length of the list. This is because the first if statement in the removeAt function checked if the index is greater or equals to the length of the list whereas it should only check if the index is greater than the length and if true, it should throw a type error.

### Describe your change:

- [x] Fix a bug or typo in the first if statement inside the removeAt() function
### Checklist:

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/JavaScript/pull/1264"></a>

### DIFF
--- a/Data-Structures/Linked-List/SinglyLinkedList.js
+++ b/Data-Structures/Linked-List/SinglyLinkedList.js
@@ -181,7 +181,7 @@ class LinkedList {
   // Removes the node at specified index
   removeAt (index) {
     // Check if index is present in list
-    if (index < 0 || index >= this.length) {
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
This can't be quite right either. It seems the code intended to use 0-based indexing. This would give us indices from `0` (first) to `length - 1` (last), which is what the code correctly checks against. The later check `index === this.length` is presumably wrong however and should be fixed to be `index === this.length - 1`.



Also, please add tests.

sandeepvinnu (approved):

PR #1266, opened 2022-11-26
algorithm: signum

SpiderMath:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

This PR is to add the implementation of the signum function. TheAlgorithms entry is [here](https://the-algorithms.com/algorithm/signum)

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/JavaScript/pull/1266"></a>

### DIFF
--- a/Maths/Signum.js
+++ b/Maths/Signum.js
@@ -1,0 +1,23 @@
+/*
+  A program to demonstrate the implementation of the signum function,
+  also known as the sign function, in JavaScript.
+
+  The signum function is an odd mathematical function, which returns the
+  sign of the provided real number.
+  It can return 3 values: 1 for values greater than zero, 0 for zero itself,
+  and -1 for values less than zero
+
+  Wikipedia: https://en.wikipedia.org/wiki/Sign_function
+*/
+
+/**
+ * @param {Number} input
+ * @returns {-1 | 0 | 1} sign of input
+ */
+function signum (input) {
+  if (input === 0) return 0
+  else if (input < 0) return -1
+  else return 1
+}
+
+export { signum }
--- a/Maths/Signum.js
+++ b/Maths/Signum.js
@@ -15,7 +15,7 @@
  * @returns {-1 | 0 | 1 | NaN} sign of input (and NaN if the input is NaN)
  */
 function signum (input) {
-  if(isNaN(input)) return NaN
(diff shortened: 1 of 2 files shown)


appgurueu (commented):
Looks fine; note: this will currently return `1` for NaN; returning NaN would be more sensible IMO.

SpiderMath:
Oh yes. I completely missed that case. I'm sorry. Thanks for pointing it out, I'll fix it right away.

appgurueu (dismissed):

Maths/Signum.js
```diff
@@ -0,0 +1,24 @@
...
+  Wikipedia: https://en.wikipedia.org/wiki/Sign_function
+*/
+
+/**
+ * @param {Number} input
+ * @returns {-1 | 0 | 1 | NaN} sign of input (and NaN if the input is NaN)
+ */
+function signum (input) {
+  if (isNaN(input)) return NaN
+  else if (input === 0) return 0
```
raklaptudirm, line 19: Else is unnecessary as you are returning in each branch.

Maths/Signum.js
```diff
@@ -0,0 +1,24 @@
...
+  else if (input < 0) return -1
+  else return 1
```
raklaptudirm, line 21: Still returns 1 for stuff like strings. I would suggest adding a conditional here and making NaN the default case.

raklaptudirm (changes requested):

Maths/Signum.js
SpiderMath: Fixed :+1:

Maths/Signum.js
raklaptudirm: @SpiderMath looks like ya forgot about this.

SpiderMath: Oops

appgurueu:
You still have an `else` in each branch. Generally using `else`-`if`s even if you are returning on each branch is not necessarily bad practice because it may help improve readability (so I'm fine with it either way), but in this particular example, your `if`s are simple enough that you should just get rid of the redundant `else`s as the `return`s already do the job.

appgurueu (approved):

raklaptudirm (approved):

PR #1273, opened 2022-12-19
--sum-of-array

Yerenzter:
I propose a built-in JavaScript function which is ```sum()``` that sum all the lengths of an array.

```
let sum = (y) => {
        let sum = 0;

        for(r=0; r<y.length; r++) {
                sum += y[r];
        }

        return sum;
};
```

### DIFF
--- a/Maths/SumOfArray.js
+++ b/Maths/SumOfArray.js
@@ -1,0 +1,11 @@
+// It will sum of how much the length of an array.
+
+let sum = (y) => {
+        let sum = 0;
+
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
* Missing tests.
* Improper formatting; poor variable naming; `r` is global.
* Missing documentation.
* Consider using `reduce` for a one-liner or alternatively using the `for`-`of` loop construct.
* Might be too trivial to warrant inclusion; not sure about that yet.

Please refer to our contribution guidelines.

Issue #1280, opened 2023-02-04
[BUG]:  DepthFirstSearch.js in trees folder is not working properly


sundaram2021:
### Description

tarverseDFS function is not working

### Expected Behavior

tarverseDFS function should traverse the tree

### Actual Behavior

In tarverseDFS function , the res array ia empty or undefined

appgurueu:
Works fine for me:

```javascript
$ node
Welcome to Node.js v16.19.0.
Type ".help" for more information.
> 
> function traverseDFS (root) {
...   const stack = [root]
...   const res = []
... 
...   while (stack.length) {
...     const curr = stack.pop()
...     res.push(curr.key)
... 
...     if (curr.right) {
...       stack.push(curr.right)
...     }
... 
...     if (curr.left) {
...       stack.push(curr.left)
...     }
...   }
... 
...   return res.reverse()
... }
undefined
> traverseDFS({key: 42, left: {key: 33}})
[ 33, 42 ]
```

That said, there seem to be multiple issues with this code:

* No tests?
* No proper module usage: Just writes its functions to the global environment (ugh)
* General graph traversal method limited to trees (binary trees even), pretty much not documented at all
* Side effects ("tests" seem to be embedded as `console.log` statements)

Feel free to rewrite / refactor it.

sundaram2021:
hey, while traversing the tree using dfs algorithm there is a common conception that should be followed ,

the argument 'tree' should be passed to the  function as an argument and then it should traverse it .

I have refactored the code and generated a PR ,

........

could you please merge that .

appgurueu:
You have added a commit to your repo. You haven't created a PR yet.

sundaram2021:
hey I have done that(created a PR) ,

there is issue with my connection

sorry for that

PR #1281, opened 2023-02-05
solves issue #1280

sundaram2021:
traverseDFS() is not working properly ,

while logging it says  [ undefined ].

and there is also an issue with searchDFS() function,

the following issue i have faced

The stack array is being used as a queue instead of a stack. A queue follows the first-in, first-out (FIFO) principle, while a stack follows the last-in, first-out (LIFO) principle. For a DFS algorithm, we need to use a stack data structure.

The code is pushing elements to the stack using the tree[node.right] and tree[node.left] expressions, but it should be using node.right and node.left directly. The tree array should not be used here.

The code is using tree[0] as the starting node for the search, but it should be using the entire tree object as the starting node.

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Trees/DepthFirstSearch.js
+++ b/Trees/DepthFirstSearch.js
@@ -1,76 +1,213 @@
 /*
- * Author: Surendra Kumar
+ * Author: Sundaram Kumar Jha
  * DFS Algorithm implementation in JavaScript
  * DFS Algorithm for traversing or searching graph data structures.
 */
 
-function traverseDFS (root) {
-  const stack = [root]
-  const res = []
+function dfs(tree) {
+  if (!tree) return;
+  console.log(tree.value);
+  for (const child of tree.children) {
+    dfs(child);
+  }
+}
 
-  while (stack.length) {
-    const curr = stack.pop()
-    res.push(curr.key)
+// This function takes a tree node as an argument, and uses recursion to traverse 
+// the tree in a depth-first manner. The function first logs the value of the current
+//  node and then iterates through its children, calling the function on each child. 
+// The function stops when it encounters a null node.
 
-    if (curr.right) {
-      stack.push(curr.right)
-    }
+function searchDFS(tree, target) {
+  if (tree.value === target) {
+    return true;
+  }
 
-    if (curr.left) {
-      stack.push(curr.left)
+  for (let i = 0; i < tree.children.length; i++) {
+    if (searchDFS(tree.children[i], target)) {
+      return true;
     }
   }
 
-  return res.reverse()
+  return false;
 }
 
-function searchDFS (tree, value) {
-  const stack = []
+// This function takes two arguments: tree, which is the tree to be searched, and target, 
+// which is the value you're searching for. The function uses a depth-first search algorithm to 
+// traverse the tree and look for the target value.
 
-  stack.push(tree[0])
+// The function starts by checking if the current node has the target value. 
+// If it does, it returns true to indicate that the target value has been found. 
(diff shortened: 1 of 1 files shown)


Trees/DepthFirstSearch.js
```diff
@@ -1,76 +1,213 @@
...
  * DFS Algorithm implementation in JavaScript
  * DFS Algorithm for traversing or searching graph data structures.
 */
 
-function traverseDFS (root) {
-  const stack = [root]
-  const res = []
+function dfs(tree) {
+  if (!tree) return;
+  console.log(tree.value);
```
appgurueu, line 9: This should either be calling a visitor function or yielding. No side effects.

Trees/DepthFirstSearch.js
```diff
@@ -1,76 +1,213 @@
...
-      stack.push(curr.right)
-    }
+function searchDFS(tree, target) {
+  if (tree.value === target) {
+    return true;
+  }
 
-    if (curr.left) {
-      stack.push(curr.left)
+  for (let i = 0; i < tree.children.length; i++) {
```
appgurueu, line 25: Just use `for (const child of tree.children)`

Trees/DepthFirstSearch.js
```diff
@@ -1,76 +1,213 @@
...
+// If it does, it returns true to indicate that the target value has been found. 
+// If not, the function loops over the children of the current node and calls itself recursively on each child. 
+// If any of the recursive calls returns true, the function also returns true to indicate that the target value 
+// has been found. If none of the recursive calls returns true, the function returns false to indicate that the 
+// target value was not found in the tree
 
-  while (stack.length !== 0) {
-    for (let i = 0; i < stack.length; i++) {
-      const node = stack.pop()
+const tree = {
```
appgurueu, line 45: This should be in a proper test case.

Trees/DepthFirstSearch.js
```diff
@@ -1,76 +1,213 @@
...
     }
   }
 
-  return res.reverse()
+  return false;
 }
 
-function searchDFS (tree, value) {
-  const stack = []
+// This function takes two arguments: tree, which is the tree to be searched, and target,
```
appgurueu, line 34: This should be in a JS doc comment

Trees/DepthFirstSearch.js
sundaram2021: But for what

Trees/DepthFirstSearch.js
sundaram2021: yeah we can use it , its a matter of preference but here it provides much more fine control

Trees/DepthFirstSearch.js
appgurueu: For reusability.

Trees/DepthFirstSearch.js
appgurueu: I don't see why the index would be needed.

sundaram2021: I have done that

Trees/DepthFirstSearch.js
sundaram2021: can you provide some more info on that  like what kind of test you want,

it will help me to make changes then....

appgurueu: See https://github.com/TheAlgorithms/TypeScript/blob/master/CONTRIBUTING.md#writing-good-tests

appgurueu:
Issue was closed via another PR.

PR #1296, opened 2023-02-20
10functions documentation

qq1054129023:
Added a document that describes functions and why the have high complexity in 10funcs_documentation.md


appgurueu (changes requested):
This seems to be a documentation-only change which greatly duplicates the existing code. Follow-up refactor PRs to reduce cyclomatic complexity may be accepted, but I'm afraid this PR is not what we're looking for.

PR #1299, opened 2023-02-22
Make ArbitraryBase Unicode-aware

lionel-rowe:
https://mathiasbynens.be/notes/javascript-unicode#counting-symbols

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Conversions/ArbitraryBase.js
+++ b/Conversions/ArbitraryBase.js
@@ -1,4 +1,17 @@
 /**
+* Divide two numbers and get the result of floor division and remainder
+* @param {number} dividend
+* @param {number} divisor
+* @returns {[result: number, remainder: number]}
+*/
+const floorDiv = (dividend, divisor) => {
+  const remainder = dividend % divisor
+  const result = Math.floor(dividend / divisor)
+
+  return [result, remainder]
+}
+
+/**
 * Converts a string from one base to other
 * @param {string} stringInBaseOne String in input base
 * @param {string} baseOneCharacters Character set for the input base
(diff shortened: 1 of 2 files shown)


appgurueu (commented):
The fix is sensible if Unicode is to be supported but it worsens the time complexity since `.at` is linear time in the size of the base string rather than constant time. Consider extracting the Unicode characters into an array as "preprocessing".

lionel-rowe:
> The fix is sensible if Unicode is to be supported but it worsens the time complexity since `.at` is linear time in the size of the base string rather than constant time. Consider extracting the Unicode characters into an array as "preprocessing".

I'm not convinced that's true — surely accessing an array index is constant time, just like accessing an object property. There's no reason for `at` to be iterating over the array, as `arr.at(n)` acts identically to `arr[n]` for non-negative integer values of `n`. Maybe I should just swap those out for square-bracket accessors as they'll never be negative anyway.

The only linear time parts I added were these:

```js
  const baseOneCharacters = [...baseOneCharacterString]
  const baseTwoCharacters = [...baseTwoCharacterString]
```

But they only run once per function call, and in any case, the first argument is already iterated prior to being reversed.

appgurueu (approved):
Nevermind, you're correct. Glancing over the changes I missed the spread operators and confused [Array.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) with [String.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/at), of which I had incorrectly assumed that it was linear time (it is not, it's constant time - it's basically just `charAt` + negative index support).

Is there any particular reason to prefer `x.at(y)` over `x[y]` here though? `y` will always be positive, so I'd prefer standard array indexing here.

Additionally, it would be good if you could convert the "test comments" into proper Jest tests.

Thanks for the fix.

raklaptudirm:
@appgurueu Should I wait for the changes you suggested or good to merge?

lionel-rowe:
Changing `baseTwoCharacters.at(remainder)` to `baseTwoCharacters[remainder]` uncovered a performance bug — `value` was often not an integer, because the division result wasn't being floored. `charAt` and `at` both silently floor their arguments, which obscured the bug:

```js
'abc'.charAt(1.5)   // b
;[...'abc'].at(1.5) // b
```

However, as a result, the loop was being iterated over many more times than necessary as `value < 1` got repeatedly divided until it reached floating-point `0`, with all the extra zero-characters finally being lopped off at the end with `` .replace(new RegExp(`^${baseTwoZero}+`), '') ``. Without that regex replacement, `convertArbitraryBase('98', '0123456789', '01234567')` had 358 leading zeros (i.e. 358 unnecessary iterations).

I've fixed that by using floor division and removing the regex replacement.

Conversions/ArbitraryBase.js
```diff
@@ -30,21 +46,11 @@ const convertArbitraryBase = (stringInBaseOne, baseOneCharacters, baseTwoCharact
   let stringInBaseTwo = ''
   const stringTwoBase = baseTwoCharacters.length
   while (value > 0) {
-    const remainder = value % stringTwoBase
-    stringInBaseTwo = baseTwoCharacters.charAt(remainder) + stringInBaseTwo
-    value /= stringTwoBase
+    const [divisionResult, remainder] = floorDiv(value, stringTwoBase)
+    stringInBaseTwo = baseTwoCharacters[remainder] + stringInBaseTwo
```
appgurueu, line 50: Wait. Taking a closer look, this is still linear time (making the entire algo quadratic time in the number of digits), since it always prepends the character, which is a linear time operation since the string has to be copied. Ideally, the digits should be pushed to an array, which should then be `reverse`d and finally `join`ed.

Note: This is just a suggestion for further improvement, your PR is good to be merged as is.

appgurueu (dismissed):

Conversions/ArbitraryBase.js
lionel-rowe: Fixed. I also added an alternate version that uses [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), which automatically uses floor division and will work for arbitrarily long input strings.

appgurueu (approved):
Excellent work.

raklaptudirm (approved):

PR #1303, opened 2023-02-27
fixes: #1280 and test:  added tests for DFS algorithm 

Ranganathan7:
…for DepthFirstSearch algorithm.

### Describe your change:

### Checklist:

### DIFF
--- a/Trees/DepthFirstSearch.js
+++ b/Trees/DepthFirstSearch.js
@@ -4,11 +4,11 @@
  * DFS Algorithm for traversing or searching graph data structures.
 */
 
-// traverses a give tree from specified root
-function traverseDFS (tree, root) {
+// traverses a give tree from specified root's value
+function traverseDFS (tree, rootValue) {
   const stack = []
   const res = []
-  stack.push(searchDFS(tree, root))
+  stack.push(searchDFS(tree, rootValue))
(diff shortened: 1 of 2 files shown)


Trees/DepthFirstSearch.js
```diff
@@ -4,35 +4,32 @@
...
 */
 
-function traverseDFS (root) {
-  const stack = [root]
+// traverses a give tree from specified root
+function traverseDFS (tree, root) {
+  const stack = []
   const res = []
-
+  stack.push(searchDFS(tree, root))
```
appgurueu, line 11: Why are you calling `searchDFS` here?

Ranganathan7: To find the node with given value. The root just has the value property, so to get the entire node with that value, searchDFS is called.

appgurueu: hmm, I think the root should be given by "reference" (i.e. index or node object) rather than by value, which requires searching for it; either way it's not clear that `root` is actually the *value* of the root

Ranganathan7: made changes in function parameter and variable naming to make it clear.

appgurueu (approved):

raklaptudirm (approved):

PR #1304, opened 2023-03-01
added mergesortV2

RAHULANUP:
Added a sorting algorithm(merge sort)...

### DIFF
--- a/Sorts/MergeSortV2.js
+++ b/Sorts/MergeSortV2.js
@@ -1,0 +1,119 @@
+/*
+-----------Merge Sort------------
+Merge sort is similar to the quick sort 
+algorithm as it uses the divide and conquer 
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
I'm afraid we have to reject this for multiple reasons:

* **It's a duplicate**. We already have [MergeSort](https://github.com/TheAlgorithms/JavaScript/blob/master/Sorts/MergeSort.js).
* If it was better than the current implementation, the current implementation should be replaced; this is however much worse
* It is implemented improperly (highly inefficiently): Merge should not be sorting the two halves using selectionsort
* It does not follow our coding style / conventions; code is duplicated
* It does not have proper JSDoc comments
* It is completely untested

PR #1315, opened 2023-04-28
feat: Ciphers/MorseCode Algorithm

mrmagic2020:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:


mrmagic2020:
probably ignore the style check of MorseCode.js, since I have to store the morse code with its corresponding string character; or there might be a better way to do this

mrmagic2020:
> > probably ignore the style check of MorseCode.js, since I have to store the morse code with its corresponding string character; or there might be a better way to do this
> 
> It's telling you that there is no need to quote these properties. You should be able to safely remove the quotes. In the case of the digits this might indeed seem dubious, but I think it plays out fine because of JS's string coercion.

Yep! Tested the new code and it worked all the same.

appgurueu (approved):

raklaptudirm (approved):

PR #1316, opened 2023-04-29
feat: new algorithm to count bits

nikolaospanagopoulos:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Bit-Manipulation/BinaryCountBitsRegex.js
+++ b/Bit-Manipulation/BinaryCountBitsRegex.js
@@ -4,18 +4,18 @@
  *I tried to think of every possible input
  **/
 
-const countBitsWithRegex = (number) => {
+const CountBitsWithRegex = (number) => {
   if (typeof number !== 'number') {
-    //because an "" is evaluated to 0
+    // because an "" is evaluated to 0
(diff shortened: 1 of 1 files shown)


appgurueu (commented):
I don't see a good reason to add this. As I see it, this is pretty much a trivial way to inefficiently count the number of set bits involving much higher level concepts and ultimatively having `toString(2)` do all the work. It might have its place in code golfing, but algorithmically, I don't see significant educational value. This is "taking a sledgehammer to crack a nut".

It has some JS educational value though.

In its current form, I see the following issues with this PR:

* No proper doc comment
* Missing tests; the same tests as for the other implementation should be applied without duplicating (simply copy-pasting) them
* All the type checks obscure the (trivial) implementation; they should probably be removed
* Not mentioning the inferiority of this implementation (granted, asymptotically both algos are linear time in the number of bits) as this impl. creates an unnecessary temporary string (e.g. using linear auxiliary space) and throws RegEx at it

Overall, I'd be leaning towards rejecting this. It might be acceptable if the changes are made. @raklaptudirm thoughts?

PR #1320, opened 2023-05-01
Update LFUCache.js

shpetimaliu:
- I think It is not necessary to use Object.seal in CacheNode and LFUCache since you are not modifying anything in them after they are created.

- In insert and refresh in FrequencyMap, you don't need to create a new Set if a specific frequency doesn't exist in the Map. It could be created immediately and would be more efficient.

- In LFUCache, instead of calling the methods cache.has(key) and cache.get(key), you can access the cache value directly using this.cache.get(key) and this.cache.has(key). This is shorter and clearer.

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Cache/LFUCache.js
+++ b/Cache/LFUCache.js
@@ -1,253 +1,146 @@
 class CacheNode {
-  constructor (key, value, frequency) {
-    this.key = key
-    this.value = value
-    this.frequency = frequency
-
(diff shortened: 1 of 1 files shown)


appgurueu:
CI is failing.

PR #1323, opened 2023-05-11
feat: add preOrder algorithm for tree data structure. test: add testc…

Aryadev19:
…ases for preOrder algortihm

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Data-Structures/Tree/PreOrder.js
+++ b/Data-Structures/Tree/PreOrder.js
@@ -1,0 +1,10 @@
+function preOrder (root) {
+  if (root === null) return []
+
+  const left = preOrder(root.left)
+  const right = preOrder(root.right)
+
+  return [root.value, ...left, ...right]
+}
+
+export { preOrder }
--- a/Data-Structures/Tree/test/PreOrder.test.js
+++ b/Data-Structures/Tree/test/PreOrder.test.js
@@ -1,0 +1,92 @@
+import { preOrder } from '../PreOrder'
+
+describe('preOrder function', () => {
+  test('should return an empty array for an empty tree', () => {
+    const tree = null
+    expect(preOrder(tree)).toEqual([])
+  })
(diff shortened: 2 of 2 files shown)


appgurueu (commented):
Implementation looks fine, apart from the fact that it has quadratic time complexity due to the constant array copying. Either fix this or add a big fat warning; I'd expect a pre-order traversal to have linear time complexity. Ideally you would use a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) instead of pushing to a list IMO.

The tests seem somewhat excessive for such a simple function (my rule of thumb roughly is 1:1 implementation to tests, 1:2 or 1:3 at most). You could also greatly shorten the tests by writing a utility constructor - something like `const T = (left, value, right) => {left: left, value: value, right: right}`.

I also think we should start consolidating (binary) tree algorithms in a (binary= tree class. A class would be a good idea to document the expected tree format and to group together related binary tree algorithms, including postorder and inorder traversal.

Finally, this is missing a doc comment; generally that would be fine for such a simple function ("self-documenting"), but I think that at least the tree format should be documented here.

Aryadev19:
You are probably right, implementing all the methods in a tree class will be better. I am totally new to contributing to open source, so I am still learning about good test cases, I have one doubt one method is missing from the tree that is tree mirroring, I was thinking about doing that, so should it be a separate algorithm like in the python one or should I just update the tree file.

appgurueu:
You should group algorithms working on the same tree structure in the same tree class. I assume you're going to implement a `BinaryTree` class?

Aryadev19:
Yeah I think there's only a BST class, not binary tree, so I can implement binary tree and implement some of the methods, I'll work on that then!

PR #1324, opened 2023-05-12
Added the optimized solution for promblem 44

utkarsh-shrivastav77:
Optimised the solution for problem 44 in o(n) time complexity.


utkarsh-shrivastav77:
Sir please check the code let me know if there is any error

appgurueu (commented):
Please adapt the tests; they can't possibly work for this, since you renamed the function.

Second, the comments are very redundant with the code (and also read as if they had been written by a certain large language model). `pentagonalIndices` isn't a set of indices, it's a set of numbers.

I also don't see how this is supposed to be "O(n)" considering it has two nested for loops which I would assume to be O(n²) given that both for loops look like they are roughly O(k). What even is n?

PR #1348, opened 2023-08-13
Update: Added Unary Operator in SumOfDigits algorithm

chetannada:
Update: I have Added Unary Operator (+) in SumOfDigits algorithm, It's Optimized Solution in current tech stack.
If there is any thing need to update in this PR , I'll update it.
Thanks.

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Maths/SumOfDigits.js
+++ b/Maths/SumOfDigits.js
@@ -8,7 +8,6 @@
 /*
   The given input is converted to a string, split into an array of characters.
   This array is reduced to a number using the method <Array>.reduce
-  NOTE: The Unary Operator (+) is used here for converting String to Number since without that it would result in a String output.
 */
 function sumOfDigitsUsingString (number) {
   if (number < 0) number = -number
--- a/Maths/SumOfDigits.js
+++ b/Maths/SumOfDigits.js
@@ -8,12 +8,12 @@
 /*
   The given input is converted to a string, split into an array of characters.
   This array is reduced to a number using the method <Array>.reduce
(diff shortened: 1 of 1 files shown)


Maths/SumOfDigits.js
```diff
@@ -8,12 +8,12 @@
 /*
   The given input is converted to a string, split into an array of characters.
   This array is reduced to a number using the method <Array>.reduce
-  NOTE: The final parseInt is just there in cases where 1 digit numbers are given, since without that it would result in a String output.
+  NOTE: The Unary Operator (+) is used here for converting String to Number since without that it would result in a String output.
```
appgurueu, line 11: I don't think `+` in JS has a purpose other than converting strings to number?

chetannada: Well, I have added above sentence `NOTE: The Unary Operator (+) is used here for converting String to Number since without that it would result in a String output.` because i have removed  `parseInt` from the below code.

`unary plus (+)` is the fastest and preferred way of converting string into a number so I have used it here. 

Do i need to remove this sentence `NOTE: The Unary Operator (+) is used here for converting String to Number since without that it would result in a String output.` ? Please confirm. Thanks.

appgurueu: Yes, I would like you to remove this sentence. The only purpose of `+` is to convert strings to numbers. It doesn't need a comment, just like `x++` doesn't need a comment "incrementing x".

raklaptudirm (approved):

PR #1351, opened 2023-08-19
Mateusmesko patch 1

mateusmesko:
Mateus Mesko Vaz

### DIFF
--- a/Maths/IsDivisible.js
+++ b/Maths/IsDivisible.js
@@ -1,23 +1,20 @@
-// Checks if a number is divisible by another number.
 function checkFiniteNumbers(num1, num2) {
   if (!Number.isFinite(num1) || !Number.isFinite(num2)) {
+    throw new TypeError('Expected a valid real number')
     throw new TypeError('Expected valid real numbers');
   }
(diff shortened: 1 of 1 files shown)


appgurueu (commented):
Frankly I don't see how this improves the code besides removing some outcommented code (which should be converted to tests). The entire "isDivisible" boils down to `a % b === 0` which doesn't really deserve its own function; it should probably be deleted entirely.

Maths/IsDivisible.js
```diff
@@ -1,15 +1,20 @@
...
   }
-  return num1 % num2 === 0
 }
 
-// isDivisible(10, 5) // returns true
-// isDivisible(123498175, 5) // returns true
-// isDivisible(99, 5) // returns false
+export const isDivisible = (num1, num2) => {
+  checkFiniteNumbers(num1, num2);
+  if(isZero(num2))
```
CarlosZoft, line 18:
```suggestion
  if(num2 !== 0)
```

CarlosZoft (changes requested):

PR #1369, opened 2023-09-23
Singly Linked List implementation using Javascript (Recursive-Approach)

Suryac72:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:


appgurueu (commented):
I don't see the benefits of a "recursive approach" here. All loops can trivially be transformed into (tail) recursion, but here that seems to just lead to harder to read, likely less performant (function call overhead) and less memory efficient (due to stack usage, assuming now TCO) code with dirtier interfaces (exposure of "internal" parameters like `currentNode`; alternative would require cluttering the code even more) here.

I also don't see how this could be merged with our existing linked list in a decent way. Of course we could just add all these methods with a `Recursive` suffix, but that would be pretty messy. Having a second "recursive linked list" which implements the same "linked list" interface would be a proper solution, but again, for the aforementioned reasons I think this offers no advantages over the existing "iterative" implementation.

This also isn't in a mergeable state right now; it's lacking tests (should use the same tests as the existing linked list), the doc comments are messy, and many functions don't make much sense (for example: why is there both a `size` field and a `lengthRecursive` method which traverses the list to find the size?).

All in all, I'm in favor of closing this PR in its current state; the use of recursion here does not seem useful to me.

@raklaptudirm thoughts?

raklaptudirm:
I agree that adding trivial recursion adds nothing of value to the algorithm. Closing since we seem to have a consensus.

Issue #1341, opened 2023-07-25
[OTHER]: why element is array in Max Heap?


GongSunJie-Plus:
### What would you like to share?

`
maxHeap.insert([4])
`

just use number

`
maxHeap.insert(4)
`

appgurueu:
The example seems to be wrong; we should just remove it. Moreover, the max heap seems redundant with our min heap. Feel free to deduplicate the code.

Bajpai25:
@appgurueu  Could you please assign me this issue .

PR #1372, opened 2023-09-26
Removed array Insertion and added numbers Fixes: #1341

ROHAN13498:

Fixes: #1341  
### Describe your change:

- [ ] Add an algorithm?
- [x] Documentation change?

### Checklist:

### DIFF
--- a/Data-Structures/Heap/MaxHeap.js
+++ b/Data-Structures/Heap/MaxHeap.js
@@ -74,12 +74,12 @@ class BinaryHeap {
 // Example
 
 // const maxHeap = new BinaryHeap()
-// maxHeap.insert([4])
-// maxHeap.insert([3])
-// maxHeap.insert([6])
-// maxHeap.insert([1])
-// maxHeap.insert([8])
-// maxHeap.insert([2])
+// maxHeap.insert(4)
+// maxHeap.insert(3)
+// maxHeap.insert(6)
+// maxHeap.insert(1)
+// maxHeap.insert(8)
+// maxHeap.insert(2)
 // const mx = maxHeap.extractMax()
 
 export { BinaryHeap }


appgurueu (approved):
That comment should probably be removed or turned into a test case. Either way this change is an improvement already though.

CarlosZoft (approved):

raklaptudirm:
@ROHAN13498 are you interested in implementing the test cases? Cause I see no point in improving upon something that is explicitly disallowed by the contribution guidelines.

ROHAN13498:
Done. I will work on it.

ROHAN13498:
> @ROHAN13498 are you interested in implementing the test cases? Cause I see no point in improving upon something that is explicitly disallowed by the contribution guidelines.


Hey,There is no testcases file for maxheap should I create a newfile and work on it?

image

raklaptudirm:
> > @ROHAN13498 are you interested in implementing the test cases? Cause I see no point in improving upon something that is explicitly disallowed by the contribution guidelines.
> 
> Hey,There is no testcases file for maxheap should I create a newfile and work on it?
> 
> image

Yes you should.

appgurueu:
The current situation is a bit awkward; we have a max heap, a min heap, and a priority queue, all of which are more or less based on similar code. This should ideally be deduplicated; max and min heap can effectively use the same code with a different comparison function, and a priority queue should just add an index from element to position in heap on top of that.

ROHAN13498:
> The current situation is a bit awkward; we have a max heap, a min heap, and a priority queue, all of which are more or less based on similar code. This should ideally be deduplicated; max and min heap can effectively use the same code with a different comparison function, and a priority queue should just add an index from element to position in heap on top of that.

So you are suggesting that Instead of having separate implementations for Max Heap and Min Heap, it's better to create a single generic heap data structure that can handle both max and min heaps by using a comparison function to determine the order of elements.

Should I work on implementing this  single heap data structure or write tests for max heap?

appgurueu:
> > The current situation is a bit awkward; we have a max heap, a min heap, and a priority queue, all of which are more or less based on similar code. This should ideally be deduplicated; max and min heap can effectively use the same code with a different comparison function, and a priority queue should just add an index from element to position in heap on top of that.
> 
> So you are suggesting that Instead of having separate implementations for Max Heap and Min Heap, it's better to create a single generic heap data structure that can handle both max and min heaps by using a comparison function to determine the order of elements.

Yes.

> Should I work on implementing this single heap data structure or write tests for max heap?

The former; the latter would pretty much be manual test duplication, which seems pretty unnecessary to me.

ROHAN13498:
> > So you are suggesting that Instead of having separate implementations for Max Heap and Min Heap, it's better to create a single generic heap data structure that can handle both max and min heaps by using a comparison function to determine the order of elements.
> 
> Yes (though my suggestion goes even further: we could deduplicate some heap and priority queue code by making the priority queue extend or use the heap, but you don't need to do this because it is trickier).
> 
> > Should I work on implementing this single heap data structure or write tests for max heap?
> 
> The former; the latter would pretty much be manual test duplication, which seems pretty unnecessary to me.

I'll start by working on test cases file for maxheap data structure  for now . Once that's in place, I'll try to tackle code deduplication.

appgurueu:
> I'll start by working on test cases file for maxheap data structure for now . Once that's in place, I'll try to tackle code deduplication.

As said, I don't think that's a good idea. You will be duplicating min heap tests effectively.

ROHAN13498:
Hey @appgurueu  I have deduplicated the codes of min heap and maxheap into single binaryheap class should I create a issue for it and create a PR or should I do PR directly.

PR #1384, opened 2023-10-01
Added BinaryToText.js and BinaryToText.test.js

User-DK:
### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Conversions/BinaryToText.js
+++ b/Conversions/BinaryToText.js
@@ -1,0 +1,25 @@
+// How to Convert Binary to Text
+// 1.Get binary byte.
+// 2.Convert binary byte to decimal.
+// 3.Get character of ASCII code from ASCII table.
+// 4.Continue with next byte.
+  
+  function binaryAgent(str) {
+    function toDecimal(x){
+        let i=0;
+        let sum=0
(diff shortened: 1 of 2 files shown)


appgurueu (changes requested):
This effectively boils down to `str.split(" ").map(x => String.fromCharCode(parseInt(x, 2))).join(" ")`. Please write it roughly like this and rename the function to a more fitting name. A single test case suffices, the second test case is not needed.

Conversions/BinaryToText.js
```diff
@@ -0,0 +1,12 @@
+// How to Convert Binary to Text
```
appgurueu, line 1: This explanation is unnecessary, the code is readable as-is. If anything, add a proper JSDoc comment. Also, format this properly please.

appgurueu:
Sorry, but we have decided that this (and PRs like this in general) don't add enough value to warrant inclusion; the split - transform - join pattern for linear time string manipulation is demonstrated elsewhere already. In the future, if in doubt whether something is algorithmically interesting enough to warrant inclusion, please open an issue.

PR #1386, opened 2023-10-01
feat : added reverse number method

dev-madhurendra:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Maths/ReverseNumber.js
+++ b/Maths/ReverseNumber.js
@@ -26,4 +26,17 @@ const ReverseNumber = (number) => {
   return reverseNumber
 }
 
-export { ReverseNumber }
+/**
+ * @author dev-madhurendra
+ * Reverses a number by converting it to a string.
+ *
+ * @param {number} n - The number to reverse.
+ * @returns {number} The reversed number.
+ *
+ * @example
+ * const reversed = reverseNumber(12345); // Returns 54321
+ */
+
+const reverseUnsigned = (n) => parseInt(n.toString().split('').reverse().join(''))
+
+export { ReverseNumber, reverseUnsigned }
--- a/Maths/test/ReverseNumber.test.js
+++ b/Maths/test/ReverseNumber.test.js
@@ -1,0 +1,20 @@
+import { ReverseNumber, reverseUnsigned } from '../ReverseNumber'
+
+describe('ReverseNumber function', () => {
+  it.each([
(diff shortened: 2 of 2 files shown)


appgurueu (approved):

raklaptudirm:
@appgurueu I think this algorithm falls in the trivial category that we discussed in the discord previously, so I don't think it will be a good addition to the repo. What is your opinion?

appgurueu:
Well it is pretty trivial, but I think it is useful to demonstrate the split-transform-join pattern for working around the immutability of strings in JavaScript.

Arguably this shouldn't be called "reverse number" though - it isn't really specific to numbers - but it should be ["reverse string"](https://github.com/TheAlgorithms/JavaScript/blob/master/String/ReverseString.js) which we already have (but not using JS's `reverse`, which would be more concise). I missed this in my review.

@dev-madhurendra please shorten the existing linear time string reversal code to a one-liner.

dev-madhurendra:
> Well it is pretty trivial, but I think it is useful to demonstrate the split-transform-join pattern for working around the immutability of strings in JavaScript.
> 
> Arguably this shouldn't be called "reverse number" though - it isn't really specific to numbers - but it should be ["reverse string"](https://github.com/TheAlgorithms/JavaScript/blob/master/String/ReverseString.js) which we already have (but not using JS's `reverse`, which would be more concise). I missed this in my review.
> 
> @dev-madhurendra please just shorten the existing linear time string reversal code to a one-liner.

@appgurueu It is already one linear 

> Well it is pretty trivial, but I think it is useful to demonstrate the split-transform-join pattern for working around the immutability of strings in JavaScript.
> 
> Arguably this shouldn't be called "reverse number" though - it isn't really specific to numbers - but it should be ["reverse string"](https://github.com/TheAlgorithms/JavaScript/blob/master/String/ReverseString.js) which we already have (but not using JS's `reverse`, which would be more concise). I missed this in my review.
> 
> @dev-madhurendra please just shorten the existing linear time string reversal code to a one-liner.

@appgurueu I think it is already one linear.

appgurueu:
It is already linear-time, but it is not a *one-liner*: It is needlessly verbose as it manually does the array reversal rather than using JS's `reverse` built-in. It should be just `str.split('').reverse().join('')` (or `[...str].reverse().join('')`).

Issue #758, opened 2021-10-08
Reduce code repetition


raklaptudirm:
We have now successfully added guidelines for importing and using algorithms in this repository, inside other algorithms. This has been done to reduce code repetition and uphold code quality. Some functions do implement their own versions of data structures and helper functions, as previously, algorithms were supposed to be completely self contained. This issue initiates a search and fix so that all functions and data structures which can be imported in an algorithm should be imported.

Also as an addition to the topic, this initiative should also weed out bad code, i.e.:
* Code which does not follow the contributing guidelines.
* Code with a bad implementation.
* Bad Code in general.
* Complicated operations without comments.

benronkth:
Hello,
We are working on some improvements to this issue. 

/Students from Royal Institute of Technology

benronkth:
There is a function called   rotate (angleInRadians) in two files, namely:
1) Data-Structures\Vectors\Vector2.js
` /**
   * Vector rotation (see https://en.wikipedia.org/wiki/Rotation_matrix)
   *
   * @param angleInRadians The angle in radians by which to rotate the vector.
   * @returns The rotated vector.
   */
  rotate (angleInRadians) {
    const ca = Math.cos(angleInRadians)
    const sa = Math.sin(angleInRadians)
    const x = ca * this.x - sa * this.y
    const y = sa * this.x + ca * this.y
    return new Vector2(x, y)
  }
  `
2) Recursive\KochSnowflake.js
`/**
   * Vector rotation (see https://en.wikipedia.org/wiki/Rotation_matrix)
   *
   * @param angleInDegrees The angle by which to rotate the vector.
   * @returns The rotated vector.
   */
  rotate (angleInDegrees) {
    const radians = angleInDegrees * Math.PI / 180
    const ca = Math.cos(radians)
    const sa = Math.sin(radians)
    const x = ca * this.x - sa * this.y
    const y = sa * this.x + ca * this.y
    return new Vector2(x, y)
  }`


As you see they should be doing the same operation but there is a conversation happening in the second file `const radians = angleInDegrees * Math.PI / 180`. Not having this line results in a test fail in Recursive\test\KochSnowflake.test.js. Which one should be kept (or renamed)?

raklaptudirm:
Since the second one is first converting to radians and then essentially running the first function, we can rewrite the second one in terms of the first.

201B153:
Is this issue open yet?

appgurueu:
[...]

PR #1388, opened 2023-10-01
fix:#758 Average Median code cleanup

developerr-ayush:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:


Maths/AverageMedian.js
```diff
@@ -9,22 +9,14 @@
  */
 
 const averageMedian = (sourceArrayOfNumbers) => {
-  let numbers = [...sourceArrayOfNumbers]
-  let median = 0
+  const numbers = [...sourceArrayOfNumbers].sort(sortNumbers)
   const numLength = numbers.length
-  numbers = numbers.sort(sortNumbers)
 
   if (numLength % 2 === 0) {
```
appgurueu, line 15: Could make this a ternary even.

Maths/AverageMedian.js
```diff
@@ -9,22 +9,14 @@
...
-
-  return median
+    return (numbers[numLength / 2 - 1] + numbers[numLength / 2]) / 2
+  } else return numbers[Math.floor(numLength / 2)]
 }
 
-const sortNumbers = (num1, num2) => {
-  return num1 - num2
-}
+const sortNumbers = (num1, num2) => num1 - num2
```
appgurueu, line 20: This way of sorting numbers is sketchy to be honest (JS doesn't have overflows, but float precision issues may result in this treating inequal numbers as equal even if float comparison would yield the correct result), but it is out of scope of this PR to fix it.

Maths/AverageMedian.js
```diff
@@ -9,22 +9,14 @@
...
 
   if (numLength % 2 === 0) {
-    median = (numbers[numLength / 2 - 1] + numbers[numLength / 2]) / 2
-  } else {
-    median = numbers[(numLength - 1) / 2]
-  }
-
-  return median
+    return (numbers[numLength / 2 - 1] + numbers[numLength / 2]) / 2
+  } else return numbers[Math.floor(numLength / 2)]
```
appgurueu, line 17: What about `numLength >> 1`?

appgurueu (approved):

Maths/AverageMedian.js
developerr-ayush: yes sure i totally forgot that we can do that way too

Maths/AverageMedian.js
developerr-ayush: let me try that.

Maths/AverageMedian.js
developerr-ayush: i can make another pr after this and can handle that

Maths/AverageMedian.js
developerr-ayush: it wont work let me explain with example 
after sort array looks like this `[1, 2, 5, 8, 9, 10, 11]`
as its length is 7 and in binary it will be 111 
shifting it two positions gives us 1 and median is at 4th position in sorted array

Maths/AverageMedian.js
developerr-ayush: we can do like this to solve that 
```js 
const averageMedian = (sourceArrayOfNumbers) => {
  const numbers = [...sourceArrayOfNumbers].sort((a, b) => sortNumbers(a, b));
  const numLength = numbers.length;

  return numLength % 2 === 0
    ? (numbers[numLength / 2 - 1] + numbers[numLength / 2]) / 2
    : numbers[Math.floor(numLength / 2)];
};

const sortNumbers = (num1, num2) => num1 - num2;
```

appgurueu (approved):
(side note: this would be called a "cleanup" rather than "optimization")

raklaptudirm (approved):

PR #1394, opened 2023-10-01
Bugfix/fix failing tests

defaude:

### Describe your change:

### Checklist:


Maths/test/ParityOutlier.test.js
```diff
@@ -9,11 +9,6 @@ describe('Testing parityOutlier function', () => {
     expect(parityOutlier([177, 5, 76, 1919])).toBe(76)
   })
 
-  it('should, if the given array has only one integer element, return the integer itself', () => {
```
appgurueu, line 12: Why is just removing the test case (rather than fixing the implementation) the correct thing to do here?

defaude: Depends on what the "correct" implementation would be! I mentioned this in the commit message already.

There is another test (see [here](https://github.com/TheAlgorithms/JavaScript/blob/master/Maths/test/ParityOutlier.test.js#L24)) that checks explicitly the inverse of this test: given just a single item, it should return `null`.

defaude: If I left the test in there and added an appropriate check to the implementation, e.g.

```js
  // if there's only one element, return it directly
  if (integers.length === 1) {
    return integers[0]
  }
```

then [this test](https://github.com/TheAlgorithms/JavaScript/blob/master/Maths/test/ParityOutlier.test.js#L24) would fail.

appgurueu (approved):

raklaptudirm (approved):

PR #1407, opened 2023-10-02
Test running overhaul, switch to Prettier & reformat everything

### LINKED ISSUE #1378: [FEATURE]: Switch to Bun or Vitest for running tests
### Motivation

As you might've heard, [Bun](https://bun.sh/) has become stable just a few weeks ago. In a nutshell, it's a drop-in replacement for "classic" Node.js that promises faster startup, overall better performance etc.

What do you guys think? Should we
- brush up & clean up the tooling overall
- switch the JS runtime to Bun
- update GitHub actions accordingly
- (optional) migrate from Jest to Bun (it comes with a Jest-compatible test runner, as well)

Please add your ideas, questions and suggestions here :)

### Examples

_No response_

### Possible workarounds

_No response_

### Additional information

_No response_

defaude:
# This PR overhauls the test setup as well as the pipeline

Fixes  #1378

## Node 20 is now the default minimum engine

This doesn't mean stuff _will_ break if people are using older versions of Node, but they will get a warning during `npm install`. Vitest should be fine with Node >= 14.18, actually. Nevertheless, it's a reasonable thing to assume everyone has at least the current "active LTS" version of Node available.

## Replacing Jest + Babel with Vitest

We already switched to ESM a while ago. There is no need to pump everything through Babel into Jest. Vitest is a blazing-fast drop-in replacement that is mostly Jest-compatible (and definitely better compatible with Jest than Bun is...). This change allows us to drop a ton of npm dependencies.

## Remove standard in favor of prettier

There was a .prettierrc file in the repo already, but for some reason we were stuck with standard. I took the liberty to clean up here, as well. It's now prettier only all the way. This means I had to automatically reformat the code once.

## Update pipeline

First of all: The long-running test we observed in #1193 is now working just fine when it's running in Vitest on Node 20 🤷

This allows us to run **all** tests in the pipeline again (and we definitely should). We should also aim for
- only ever merge any PRs when their pipeline is green
- only ever commit & push when the tests are successful (this **should** be ensured by the pre-commit hook, but I get the feeling that many people are simply suppressing them?)


.prettierrc
```diff
@@ -1,9 +1,9 @@
 {
   "arrowParens": "always",
   "bracketSpacing": true,
-  "endOfLine": "lf",
+  "endOfLine": "auto",
   "insertPragma": false,
-  "printWidth": 80,
+  "printWidth": 120,
```
appgurueu, line 6: Why?

String/test/LengthofLongestSubstringWithoutRepetition.test.js
```diff
@@ -12,15 +12,15 @@ describe('LengthOfLongestSubstring', () => {
     expect(lengthOfLongestSubstring('bbbbb')).toBe(1)
     expect(lengthOfLongestSubstring('pwwkew')).toBe(3)
     expect(lengthOfLongestSubstring(' ')).toBe(1)
-    expect(lengthOfLongestSubstring('abcdefghijklmnaaaaa')).toBe(13)
+    expect(lengthOfLongestSubstring('abcdefghijklmnaaaaa')).toBe(14)
```
appgurueu, line 15: That's odd, how did this pass our CI previously?

appgurueu (approved):
Looks good to me, though I'd prefer keeping line endings at LF in the prettier config unless there is a good reason against it. I also wonder how our CI apparently passed with failing tests (but analyzing or fixing that can be done later, it doesn't have to be in this PR).

Note to self & Rak: Look at the commits individually to not get caught up in the large formatting diff, *don't squash*.

appgurueu (approved):

.prettierrc
```diff
@@ -1,9 +1,9 @@
 {
   "arrowParens": "always",
   "bracketSpacing": true,
-  "endOfLine": "lf",
+  "endOfLine": "auto",
```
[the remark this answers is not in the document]
defaude: That should be fixed by git automatically, no?

But I can fall back to the previous setting if you prefer :)

.prettierrc
defaude: Personal preference, to be honest. On today's screens which are mostly widescreen, just 80 feels a bit limiting. I've been working with 120 in most projects for years now :)

But also here: If you prefer 80, I'll gladly revert back to it.

String/test/LengthofLongestSubstringWithoutRepetition.test.js
defaude: It didn't :)

See https://github.com/TheAlgorithms/JavaScript/actions/runs/6383041840/job/17322888064

.prettierrc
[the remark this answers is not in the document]
raklaptudirm: Well it "should" be but having a backup system is never a bad idea.

String/test/LengthofLongestSubstringWithoutRepetition.test.js
raklaptudirm: Merged without CI passing? Weird, GitHub should prevent me from doing that.

raklaptudirm: Bruh why was it marked as passed?
Screenshot 2023-10-03 at 12 14 12 PM

.github/workflows/Ci.yml
```diff
@@ -15,20 +15,15 @@ jobs:
...
           cache: npm
 
       - name: 📦 Install dependencies
         run: npm ci
 
       - name: 🧪 Run all tests
-        if: ${{ github.event_name == 'push' }}
         run: npm run test
 
-      - name: 🧪 Run tests for changed files only
```
raklaptudirm, line 28: Isn't it a good idea to run the tests for the changed files only? Or is this not compatible?

String/test/LengthofLongestSubstringWithoutRepetition.test.js
defaude: I guess this is an artifact of running tests only on changed files. With my PR here, this won't happen again :)

.github/workflows/Ci.yml
defaude: Ideally, you should **always** run **every** test - exactly because of situations like [this](https://github.com/TheAlgorithms/JavaScript/pull/1407#discussion_r1343580070).
And yeah, this means that people should pay some attention that their tests not only run successfully, but do so in a timely manner.

.prettierrc
defaude: Just updated the PR with `printWidth: 80` again.

.prettierrc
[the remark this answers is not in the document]
defaude: I've set the `endOfLine` back to `"lf"` and applied the style again.

.prettierrc
appgurueu: Well, I'm not that opposed to large line limits; I have some myself. Esp. for diff viewing a lower line length limit is more comfortable though.

String/test/LengthofLongestSubstringWithoutRepetition.test.js
appgurueu: This will ideally also prevent causing regressions in dependents, so something like changing the quick sort pivot selection strategy breaking quick select shouldn't go unnoticed in the future :)

appgurueu (approved):

raklaptudirm (approved):

PR #1419, opened 2023-10-03
Feat: Min Max using Bit-Manipulation function created with test cases

nikgautamgithub:
### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Bit-Manipulation/minMaxElement.js
+++ b/Bit-Manipulation/minMaxElement.js
@@ -1,0 +1,17 @@
+/**
+ * @author : nikgautamgithub
+ *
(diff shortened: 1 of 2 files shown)


appgurueu (changes requested):
The doc comment in its current form needs to be fixed (copied to min and edited; it will only apply to `max` currently). It should also be made clear that there is a limit (32-bit ints) on the numbers, and perhaps that this is extremely likely to be slower than just `Math.min` / `Math.max`. More tests would also be good, though the logic checks out.

PR #1432, opened 2023-10-04
Create WiggleSort.js

iMeet07:
Added Wiggle Sort


appgurueu (changes requested):
Per our conventions, the tests go in a separate file. Here, that would be `Sorts/test/WiggleSort.test.js`. I have looked at the other wiggle sort and while it seems to solve the same problem, it uses a different implementation, so having both around is appropriate. You could try to deduplicate the tests though.

PR #1434, opened 2023-10-04
Sleep Sort

iMeet07:
Added Sleep Sort Algorithm


appgurueu (changes requested):
Doesn't work:

```javascript
> await sleepSort([3, 2, 1])
[ 3, 2, 1 ]
```

The problem is that you sequentially `await` the sleeping for each element, so you're not doing any sorting, just sleeping. You need to kick off the sleeping for all elements at the same time, then wait for all of them to finish. You can use `Promise.all` for this.

Sorts/SleepSort.js
```diff
@@ -21,24 +21,32 @@
...
+        sortedArray.push(num);
+        resolve();
+      }, num);
+    }));
   }
 
-  
\ No newline at end of file
+  await Promise.all(promises);
+  return sortedArray.sort((a, b) => a - b);
```
appgurueu, line 47: This doesn't belong here.

package-lock.json
```diff
@@ -11,1939 +11,7317 @@
       "devDependencies": {
         "globby": "^13.2.2",
         "husky": "^8.0.3",
+        "jest": "^29.7.0",
```
appgurueu, line 14: These changes don't belong in this PR.

package.json
```diff
@@ -14,6 +14,7 @@
   "devDependencies": {
     "globby": "^13.2.2",
     "husky": "^8.0.3",
+    "jest": "^29.7.0",
```
appgurueu, line 17: We don't use Jest anymore, we use Vitest now.

Sorts/SleepSort.js
iMeet07: Thanks for the Help

package.json
iMeet07: Ok. Got it

appgurueu (changes requested):
The changes to `package-lock.json` and `package.json` need to be removed. There is absolutely no need to import `expect` from chai.

PR #1435, opened 2023-10-04
Divisors.js

ZainabMalik5:
I have added a file for computing all unique divisors of a number in javascript in Maths folder. Please merge my PR.


ZainabMalik5:
I have made the necessary changes and sent you a pull request. Please merge
it.

ZainabMalik5:
Please merge PR #1435 if it's fine.

gaurovgiri (approved):
Great Work! Keep it up!

appgurueu (changes requested):
Missing tests. `@return{Vector}` should be `@return{[]Integer}` (`Integer` is treated as an undefined custom type; we could use `number`, but that doesn't convey the same information. Ideally consider `@typedef`ing Integer). Not sure how useful this is given that we already have prime factorization. (This gives you all divisors of a number, but I think that is almost always less useful than prime factorization.)

ZainabMalik5:
I have made the necessary changes in Divisors.js. Please review it.

Maths/Divisors.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @function getDivisors
+ * @description Generate all unique divisors of a number.
+ * @param {Integer} num
+ * @return{[]Integer} divisors
+ * @example{10-1,2,5,10}
```
appgurueu, line 6: This example is broken

appgurueu:
> I have made the necessary changes in Divisors.js. Please review it.

It's still missing tests.

PR #1437, opened 2023-10-04
Added string to fetch integer algorithm

Shantanu-Meta:
Problem statement: 
Implement the myAtoi(string s) function, which converts a string to a 32-bit signed integer (similar to C/C++'s atoi function).

The algorithm for myAtoi(string s) is as follows:

Read in and ignore any leading whitespace.
Check if the next character (if not already at the end of the string) is '-' or '+'. Read this character in if it is either. This determines if the final result is negative or positive respectively. Assume the result is positive if neither is present.
Read in next the characters until the next non-digit character or the end of the input is reached. The rest of the string is ignored.
Convert these digits into an integer (i.e. "123" -> 123, "0032" -> 32). If no digits were read, then the integer is 0. Change the sign as necessary (from step 2).
If the integer is out of the 32-bit signed integer range [-231, 231 - 1], then clamp the integer so that it remains in the range. Specifically, integers less than -231 should be clamped to -231, and integers greater than 231 - 1 should be clamped to 231 - 1.
Return the integer as the final result.
##
Input: s = "    -4193 with words"
Output:  -4193
##

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/String/stringToInt.js
+++ b/String/stringToInt.js
@@ -1,0 +1,22 @@
+var myAtoi = function(s) {
+    s = s.trim();  // Remove leading and trailing whitespace
+    if (s.length === 0) {
+        return 0;  // Handle empty string case
+    }
+    let num = 0;
+    let i = 0;
+    let sign = 1;  // 1 for positive, -1 for negative
+    if (s[i] === '+') {
+        i++;
+    } else if (s[i] === '-') {
+        i++;
+        sign = -1;
+    }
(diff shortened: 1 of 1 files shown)


Shantanu-Meta:
Hey I have made this pr to implement a new algorithm under HACKTOBERFEST2k23. Should it need any improvement ?

PR #1439, opened 2023-10-05
feat: add decimal to fraction algorithm

MarufHasan24:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Maths/Fraction.js
+++ b/Maths/Fraction.js
@@ -14,10 +14,11 @@
  * @example fraction(0.33,2) // [10, 3] 
  */
 function fraction(number, accuracy = 6) {
+    let returnVal = [];
     let inp = typeof number === "number" ? number : () => {throw new TypeError("Invalid number, a number type value expected");};
     let acc = typeof accuracy === "number" && accuracy >= 1 && accuracy <= 16 ? accuracy : () => {throw new TypeError("Invalid accuracy, a integer type value expected between 1 - 16");}
       if (Number.isInteger(inp)) {
-        return [inp, 1];
+        returnVal = [inp, 1];
       } else {
         inp = inp.toString();
         let len;
@@ -53,8 +54,9 @@ function fraction(number, accuracy = 6) {
             d = false;
           }
         }
-        return [inp, len];
+        returnVal= [inp, len];
       }
+    return returnVal;
   }
   
   export { fraction };
--- a/Maths/Fraction.js
+++ b/Maths/Fraction.js
@@ -14,49 +14,54 @@
  * @example fraction(0.33,2) // [10, 3] 
  */
 function fraction(number, accuracy = 6) {
-    let returnVal = [];
-    let inp = typeof number === "number" ? number : () => {throw new TypeError("Invalid number, a number type value expected");};
-    let acc = typeof accuracy === "number" && accuracy >= 1 && accuracy <= 16 ? accuracy : () => {throw new TypeError("Invalid accuracy, a integer type value expected between 1 - 16");}
-      if (Number.isInteger(inp)) {
-        returnVal = [inp, 1];
-      } else {
-        inp = inp.toString();
+    if (typeof number === "number" && Number.isNaN(number) && Number.isFinite(number)
+        && typeof accuracy === "number" && Number.isNaN(accuracy) && accuracy >= 1 && accuracy <= 16) {
+        let neg = 1;
+        // if number is negative then following code will run
+        if (number < 0) {
+            neg = -1;
+            number = Math.abs(number);
+        }
+        // if number is 0 then it will return [0, 1]
+        if (number === 0) return [0, 1];
+        if (Number.isInteger(number)) return [neg * number, 1];
+        // if number is not an integer then follwing code will run
+        number = number.toString();
         let len;
-        let reg = inp.match(/(\d+?)\1+$/);
-        if (reg && reg[0].length > acc) {
-          let pos = inp.split(".");
-          inp = inp.replace(reg[0], reg[1]);
-          let rec = pos[0] + pos[1].replace(reg[0], "");
-          inp = Number(rec + reg[1]) - Number(rec);
-          len = Number(
-            "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
-          );
+        let reg = number.match(/(\d+?)\1+$/);
+        // if number is repeating decimal then following code will run
+        if (reg && reg[0].length > accuracy) {
+            let pos = number.split(".");
+            number = number.replace(reg[0], reg[1]);
+            let rec = pos[0] + pos[1].replace(reg[0], "");
+            number = Number(rec + reg[1]) - Number(rec);
+            len = Number(
+                "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
+            );
         } else {
-          inp = inp.replace(".", "");
-          len = 10 ** (inp.length - 1);
-          inp = Number(inp);
+            // if number is not repeating decimal then following code will run
+            number = number.replace(".", "");
+            len = 10 ** (number.length - 1);
+            number = Number(number);
         }
-        let d = true;
-        while (d) {
-          if (inp % 5 == 0 && len % 5 == 0) {
-            inp /= 5;
-            len /= 5;
-          } else if (inp % 2 == 0 && len % 2 == 0) {
-            inp /= 2;
-            len /= 2;
-          } else if (inp % 3 == 0 && len % 3 == 0) {
-            inp /= 3;
-            len /= 3;
-          } else if (inp % 7 == 0 && len % 7 == 0) {
-            inp /= 7;
-            len /= 7;
-          } else {
-            d = false;
-          }
-        }
-        returnVal= [inp, len];
-      }
-    return returnVal;
-  }
-  
-  export { fraction };
+        // it will findout the gcd of number and len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
+        let div = gcd(number, len);
+        number /= div;
+        len /= div;
+        return [neg * number, len];
+    } else {
+        if (typeof number !== "number") throw new TypeError("Invalid number, a number type value expected");
+        if (typeof accuracy !== "number") throw new TypeError("Invalid accuracy, a number type value expected");
+        if (Number.isNaN(number)) throw new TypeError("Invalid number, a number type value expected");
+        if (Number.isNaN(accuracy)) throw new TypeError("Invalid accuracy, a number type value expected");
+        if (!Number.isFinite(number)) throw new RangeError("Invalid number, a finite number expected");
+        if (accuracy < 1 || accuracy > 16) throw new RangeError("Invalid accuracy, a integer type value expected between 1 and 16");
+    }
+}
+
+function gcd(a, b) {
+    if (b == 0) return a;
+    return gcd(b, a % b);
+}
+
+export { fraction };
--- a/Maths/Fraction.js
+++ b/Maths/Fraction.js
@@ -17,19 +17,19 @@ function fraction(number, accuracy = 6) {
     if (typeof number === "number" && Number.isNaN(number) && Number.isFinite(number)
         && typeof accuracy === "number" && Number.isNaN(accuracy) && accuracy >= 1 && accuracy <= 16) {
         let neg = 1;
-        // if number is negative then following code will run
+        // if number is a negative then following code will run
         if (number < 0) {
             neg = -1;
             number = Math.abs(number);
         }
-        // if number is 0 then it will return [0, 1]
+        // if number is a 0 then it will return [0, 1]
         if (number === 0) return [0, 1];
         if (Number.isInteger(number)) return [neg * number, 1];
-        // if number is not an integer then follwing code will run
+        // if number is a not an integer then follwing code will run
         number = number.toString();
         let len;
         let reg = number.match(/(\d+?)\1+$/);
-        // if number is repeating decimal then following code will run
+        // if number is a repeating decimal then following code will run
         if (reg && reg[0].length > accuracy) {
             let pos = number.split(".");
             number = number.replace(reg[0], reg[1]);
@@ -39,12 +39,12 @@ function fraction(number, accuracy = 6) {
                 "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
             );
         } else {
-            // if number is not repeating decimal then following code will run
+            // if number is not a repeating decimal then following code will run
             number = number.replace(".", "");
             len = 10 ** (number.length - 1);
             number = Number(number);
         }
-        // it will findout the gcd of number and len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
+        // it will find out the gcd of the number and the len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
         let div = gcd(number, len);
         number /= div;
         len /= div;
--- a/Maths/Fraction.js
+++ b/Maths/Fraction.js
@@ -1,0 +1,66 @@
+/**
+ * @function fraction
+ * @description This function returns the fraction of a float type number.
+ * @param {number} num - a float type number is expected, but an integer will also work.
+ * @param {number} accuracy - the accuracy of the fraction, the default is 6.
+ * @return {Array} - an array containing the numerator and denominator of the fraction.
+ * @see https://en.wikipedia.org/wiki/Repeating_decimal and @see https://en.wikipedia.org/wiki/Decimal#Decimal_fractions
+ * @example fraction(0.5) // [1, 2]
+ * @example fraction(0.3333333333333333) // [1, 3]
+ * @example fraction(0.25) // [1, 4]
+ * @example fraction(5.56) // [139, 25]
+ * @example fraction(0.33) // [33, 100]
+ */
+function fraction(number, accuracy = 6) {
+  let inp = typeof number === "number" ? number : false;
+  let acc = typeof accuracy === "number" ? accuracy : false;
+  if (inp !== false && acc !== false) {
+    if (Number.isInteger(inp)) {
+      return [inp, 1];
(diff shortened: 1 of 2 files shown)


RRMovies (approved):

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+ * @see https://en.wikipedia.org/wiki/Repeating_decimal and 
+ * @see https://en.wikipedia.org/wiki/Decimal#Decimal_fractions
+ * @example fraction(0.5) // [1, 2]
+ * @example fraction(0.3333333333333333) // [1, 3]
+ * @example fraction(0.25) // [1, 4]
+ * @example fraction(5.56) // [139, 25]
+ * @example fraction(0.33) // [33, 100]
+ */
+function fraction(number, accuracy = 6) {
+  let inp = typeof number === "number" ? number : false;
```
appgurueu, line 16: You should just do `if (typeof number !== "number") throw "invalid number";` right away here. Same for the accuracy. Would also save you most of the `if`-`else` zoo below.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+  let acc = typeof accuracy === "number" ? accuracy : false;
+  if (inp !== false && acc !== false) {
+    if (Number.isInteger(inp)) {
+      return [inp, 1];
+    } else {
```
appgurueu, line 21: This `else` is unneded due to the `return`.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+      inp = inp.toString();
+      let len;
+      let reg = inp.match(/(\d+?)\1+$/);
```
appgurueu, line 24: This string processing is awfully error prone (for example, what if JS decides to format your number using "nerd notation"?) and also hard to read. Replace it with proper number operations. For example, `Math.floor(number)` will be the integral part, and `number % 1` will be the fractional part.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        len = Number(
+          "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
+        );
+      } else {
+        inp = inp.replace(".", "");
+        len = 10 ** (inp.length - 1);
+        inp = Number(inp);
+      }
+      let d = true;
+      while (d) {
```
appgurueu, line 39: This logic doesn't make much sense. It seems like a hack to handle multiples of 2, 3, 5, and 7 "better"? Either way this needs an explanation.

appgurueu (commented):
Please make clear what the spec of this is. What is the accuracy? If it's in decimal places, why do you need any complicated logic beyond rounding, and choosing the appropriate power of ten for the denominator?

Maths/Fraction.js
MarufHasan24: no, @appgurueu look closely. Here if the input is an integer, most of the case which is not, will return [inp,1] else it will do it's real job. It will not stop in if, if the number is not an integer.

Maths/Fraction.js
MarufHasan24: I got your point. But here is a but here. there are 2 types of rational numbers.
- Repetitive rational number. like: 1/3 = 0.333333...
- Non-Terminating rational number. like: 5/4 = 1.25

now your idea works for the second case easily, but for terminating numbers sometimes it doesn't work like that. like : 1.235235235... though it's a terminating rational number, but function returns 1235235235/1000000000 where it should return 1234/999. here regex decides that the number is terminating or not.

Maths/Fraction.js
MarufHasan24: Here I divide the numbers by some prime numbers every time, to get the smallest numerator and denominator. you can ask why only 2,3,5 and 7. Because, there is a unlimited count of prime numbers. In school, we learnt to divide using 2,3 and 5. Because that's how we get smallest numerator and denominator. I just added 7 here, nothing else.

MarufHasan24 (commented):
Maybe I have answered all your questions, and resolved those, what I think I should.

Maths/Fraction.js
appgurueu: So the purpose of this is to shorten the fraction? If so, then this should simply divide both numerator and denominator by their GCD. No need to try out a set of primes.

Maths/Fraction.js
appgurueu: There are periodic numbers, yes. But given a finite number, you can't know whether it is supposed to be periodic, and you shouldn't decide it arbitrarily.

Maths/Fraction.js
appgurueu: I did not ask you to resolve this by making the `else` "necessary" by removing the early return and replacing it with assignment to a return value. That just makes the code even harder to read.

appgurueu:
I'd like to reiterate this:

> Please make clear what the spec of this is. What is the accuracy? If it's in decimal places, why do you need any complicated logic beyond rounding, and choosing the appropriate power of ten for the denominator?

Code-wise, it seems you misunderstood my point. I'm asking for a "flattening" of the code using early returns / throws. The logic should be something like "if input is invalid, throw. if input is trivial (integer), return early. now do the actual meaty logic (if we got here, we didn't return or throw already)".

Maths/Fraction.js
MarufHasan24: @appgurueu Yah! that's a great idea. should I add that, or you will!?

Maths/Fraction.js
MarufHasan24: here else is "necessary"! Because if the input is not a integer type number, then what will happen!

MarufHasan24:
> I'd like to reiterate this:
> 
> > Please make clear what the spec of this is. What is the accuracy? If it's in decimal places, why do you need any complicated logic beyond rounding, and choosing the appropriate power of ten for the denominator?
> 
> Code-wise, it seems you misunderstood my point. I'm asking for a "flattening" of the code using early returns / throws. The logic should be something like "if input is invalid, throw. if input is trivial (integer), return early. now do the actual meaty logic (if we got here, we didn't return or throw already)".

If we got the number as an integer, then what!? Or if the is input is not a number even, then I will wait for what? Please clear me.😟

appgurueu:
> If we got the number which is not an integer, then what!? Or if the is input is not a number even, then I will wait for what? Please clear me.😟

The logic should stay the same, it's just about how you write it. Currently the code is a mess of `if`-`else`s - e.g., `else`s to throw in error cases. What I'm proposing is to use *early `return`s and `throw`s* to clean it up structure-wise. Suppose you have

```javascript
function decimalToFraction(decimal) {
    if (typeof decimal === "number") {
        if (Number.isInteger(decimal)) {
             return [decimal, 1]
        } else {
             // this is where the logic for converting a valid decimal that isn't an integer to a fraction goes...
        }
    } else {
        throw "invalid type"
    }
}
```

then this would be better written as

```javascript
function decimalToFraction(decimal) {
    if (typeof decimal !== "number") throw "invalid type"
    if (Number.isInteger(decimal)) return [decimal, 1]
    // this is where the logic for converting a valid decimal that isn't an integer to a fraction goes...
}
```

MarufHasan24:
@appgurueu fixed all issues you have mentioned.

MarufHasan24:
@appgurueu any progress?!

raklaptudirm:
@appgurueu waiting for your updated review.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @function fraction
+ * @description This function returns the fraction of a float type number.
+ * @param {number} num - a float type number is expected, but an integer will also work.
+ * @param {number} accuracy - the accuracy of the fraction, the default is 6. like if the accuracy is 2 then for 1.333 result will be 10/9, where for 6 it will return 1333/1000.
```
appgurueu, line 5: Please formulate this mathematically - something like "fraction - num <= 10^-accuracy".

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+ * @return {Array} - an array containing the numerator and denominator of the fraction.
+ * @see https://en.wikipedia.org/wiki/Repeating_decimal and 
+ * @see https://en.wikipedia.org/wiki/Decimal#Decimal_fractions
+ * @example fraction(0.5) // [1, 2]
+ * @example fraction(0.3333333333333333) // [1, 3]
```
appgurueu, line 10: Are all these examples really useful? `0.5` and `0.25` for example seem redundant.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+ * @example fraction(0.25) // [1, 4]
+ * @example fraction(5.56) // [139, 25]
+ * @example fraction(0.33) // [33, 100]
+ * @example fraction(0.33,2) // [10, 3] 
+ */
+function fraction(number, accuracy = 6) {
+    if (typeof number === "number" && !Number.isNaN(number) && Number.isFinite(number)
```
appgurueu, line 17: You're still not using guards here; these checks are all redundant with the if-throws in the `else`. You just need to move the if-throws to the top of the function, making them guards.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            len = 10 ** (number.length - 1);
+            number = Number(number);
+        }
+        // it will find out the gcd of the number and the len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
+        let div = gcd(number, len);
+        number /= div;
+        len /= div;
+        return [neg * number, len];
+    } else {
+        if (typeof number !== "number") throw new TypeError("Invalid number, a number type value expected");
```
appgurueu, line 53: These ["guards"](https://deviq.com/design-patterns/guard-clause) should all be moved to the top of the function. They could also be simplified a bit.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        if (typeof accuracy !== "number") throw new TypeError("Invalid accuracy, a number type value expected");
+        if (Number.isNaN(number)) throw new TypeError("Invalid number, a number type value expected");
```
appgurueu, line 55: The NaN check is redundant with the negated `isFinite` check.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            number = Number(number);
+        }
+        // it will find out the gcd of the number and the len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
+        let div = gcd(number, len);
+        number /= div;
+        len /= div;
+        return [neg * number, len];
+    } else {
+        if (typeof number !== "number") throw new TypeError("Invalid number, a number type value expected");
+        if (typeof accuracy !== "number") throw new TypeError("Invalid accuracy, a number type value expected");
```
appgurueu, line 54: Redundant with the negated `isFinite` check.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        if (Number.isNaN(number)) throw new TypeError("Invalid number, a number type value expected");
+        if (Number.isNaN(accuracy)) throw new TypeError("Invalid accuracy, a number type value expected");
+        if (!Number.isFinite(number)) throw new RangeError("Invalid number, a finite number expected");
```
appgurueu, line 57: Keep only this check for `number`. Similar for `accuracy`: Keep only the range + type check, which you can do in a single `if` statement.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        if (accuracy < 1 || accuracy > 16) throw new RangeError("Invalid accuracy, a integer type value expected between 1 and 16");
+    }
+}
+
+function gcd(a, b) {
```
appgurueu, line 62: We have a gcd implementation in this repo. Import and reuse it.

Maths/test/Fraction.test.js
```diff
@@ -0,0 +1,31 @@
+import { fraction } from './../Fraction.js';
+
+describe('Fraction', () => {
+    it('should return [1, 2] for 0.5', () => {
```
appgurueu, line 4: These tests should use `it.each`.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+ * @example fraction(0.25) // [1, 4]
+ * @example fraction(5.56) // [139, 25]
+ * @example fraction(0.33) // [33, 100]
+ * @example fraction(0.33,2) // [10, 3] 
+ */
+function fraction(number, accuracy = 6) {
+    if (typeof number === "number" && !Number.isNaN(number) && Number.isFinite(number)
+        && typeof accuracy === "number" && !Number.isNaN(accuracy) && accuracy >= 1 && accuracy <= 16) {
+        let neg = 1;
+        // if number is a negative then following code will run
```
appgurueu, line 20: Redundant comment.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+ * @example fraction(0.3333333333333333) // [1, 3]
+ * @example fraction(0.25) // [1, 4]
+ * @example fraction(5.56) // [139, 25]
+ * @example fraction(0.33) // [33, 100]
+ * @example fraction(0.33,2) // [10, 3] 
+ */
+function fraction(number, accuracy = 6) {
+    if (typeof number === "number" && !Number.isNaN(number) && Number.isFinite(number)
+        && typeof accuracy === "number" && !Number.isNaN(accuracy) && accuracy >= 1 && accuracy <= 16) {
+        let neg = 1;
```
appgurueu, line 19: I'd call this sign. Also, I'd just do `const sign = Math.sign(number)` and `const abs = Math.abs(number)` here.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        // if number is a negative then following code will run
+        if (number < 0) {
+            neg = -1;
+            number = Math.abs(number);
+        }
+        // if number is a 0 then it will return [0, 1]
```
appgurueu, line 25: Redundant comment.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        if (number === 0) return [0, 1];
+        if (Number.isInteger(number)) return [neg * number, 1];
+        // if number is a not an integer then following code will run
```
appgurueu, line 28: Redundant comment.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            len = Number(
+                "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
+            );
+        } else {
+            // if number is not a repeating decimal then following code will run
+            number = number.replace(".", "");
+            len = 10 ** (number.length - 1);
+            number = Number(number);
+        }
+        // it will find out the gcd of the number and the len to reduce the fraction nomitor and denominator like 4/8 will be 1/2
```
appgurueu, line 47: This is called "shortening" the fraction. `number` and `len` are odd names for "numerator" and "denominator".

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+        if (reg && reg[0].length > accuracy) {
+            let pos = number.split(".");
+            number = number.replace(reg[0], reg[1]);
+            let rec = pos[0] + pos[1].replace(reg[0], "");
+            number = Number(rec + reg[1]) - Number(rec);
+            len = Number(
+                "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
+            );
+        } else {
+            // if number is not a repeating decimal then following code will run
```
appgurueu, line 42: Again a redundant comment.

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            number = number.replace(".", "");
+            len = 10 ** (number.length - 1);
```
appgurueu, line 44: It is not apparent to me why this is correct: Isn't `number.length` the length of the whole number (without the period, in digits), not just the "fractional" part after the period here?

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            let pos = number.split(".");
+            number = number.replace(reg[0], reg[1]);
+            let rec = pos[0] + pos[1].replace(reg[0], "");
+            number = Number(rec + reg[1]) - Number(rec);
+            len = Number(
+                "9".repeat(reg[1].length) + "0".repeat(rec.length - pos[0].length)
+            );
+        } else {
+            // if number is not a repeating decimal then following code will run
+            number = number.replace(".", "");
```
appgurueu, line 43: String manipulation for manipulating numbers is dirty here. Just multiply your numerator by the denominator, then round. Shouldn't the denominator be determined by the accuracy?

Maths/Fraction.js
```diff
@@ -0,0 +1,67 @@
...
+            neg = -1;
+            number = Math.abs(number);
+        }
+        // if number is a 0 then it will return [0, 1]
+        if (number === 0) return [0, 1];
+        if (Number.isInteger(number)) return [neg * number, 1];
+        // if number is a not an integer then following code will run
+        number = number.toString();
+        let len;
+        let reg = number.match(/(\d+?)\1+$/);
```
appgurueu, line 31: This is a hacky way to deal with "repeating" fractional parts, and I'd argue that it solves the wrong problem. Since we're dealing with finite-size numbers (floats) here, there *are* no "repeating" decimals; we can't distinguish a "repeating" decimal from a "deliberately ending decimal". We need to treat both the same (it seems "accuracy" is more of a "threshold" of when you consider a decimal repeating?).

In fact, all floats can even be represented *exactly* as fractions (though perhaps maybe not as fractions of floats, that might get hairy towards some edge cases) where the denominator is a power of two and the numerator is an integer.

I wouldn't exclude the possibility that some special provisions for repeating decimals may be useful, but then what problem this exactly solves (and how) and what guarantees it makes should be explicitly stated.

appgurueu (changes requested):
I don't see this robustly implementing (or even *precisely* defining) the problem yet; the code is also currently still quite dirty.

PR #1441, opened 2023-10-05
feat: add algorithm to evaluate postfix string

gaurovgiri:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Data-Structures/Stack/EvaluateExpression.js
+++ b/Data-Structures/Stack/EvaluateExpression.js
@@ -1,0 +1,65 @@
+/**
+ * Evaluate a numeric operations string using a stack.
+ * Supports basic arithmetic operations: +, -, *, /
+ *
+ * @param {string} expression - Numeric operations expression to evaluate.
+ * @returns {number|null} - Result of the expression evaluation, or null if the expression is invalid.
+ */
+function evaluateExpression(expression) {
+    const stack = [];
+  
+    // Helper function to perform an operation and push the result to the stack
+    function performOperation(operator) {
+      const operand2 = stack.pop();
+      const operand1 = stack.pop();
+      if (operator === '+' || operator === '-' || operator === '*' || operator === '/') {
+        if (operand1 === undefined || operand2 === undefined) {
+          return null; // Invalid expression
+        }
+        switch (operator) {
+          case '+':
+            stack.push(operand1 + operand2);
+            break;
+          case '-':
+            stack.push(operand1 - operand2);
+            break;
+          case '*':
+            stack.push(operand1 * operand2);
+            break;
+          case '/':
+            if (operand2 === 0) {
+              return null; // Division by zero
+            }
+            stack.push(operand1 / operand2);
+            break;
+        }
+      } else {
+        return null; // Invalid operator
+      }
+    }
+  
+    const tokens = expression.split(' ');
+  
+    for (const token of tokens) {
+      if (!isNaN(parseFloat(token))) {
+        // If the token is a number, push it to the stack
+        stack.push(parseFloat(token));
+      } else {
+        // If the token is an operator, perform the operation
+        const result = performOperation(token);
+        if (result === null) {
+          return null; // Invalid expression
+        }
+      }
+    }
+  
+    if (stack.length === 1) {
+      // The final result should be on the stack
+      return stack[0];
+    } else {
+      return null; // Invalid expression
+    }
+  }
+  
+  export { evaluateExpression };
+  
--- a/Data-Structures/Stack/EvaluateExpression.js
+++ b/Data-Structures/Stack/EvaluateExpression.js
@@ -8,7 +8,7 @@
 function evaluatePostfixExpression(expression) {
   const stack = [];
 
-  // Helper function to perform an operation and push the result to the stack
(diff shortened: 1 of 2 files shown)


aayush105 (approved):
The code looks good to merge.

Data-Structures/Stack/EvaluateExpression.js
```diff
@@ -0,0 +1,66 @@
...
+  
+    const tokens = expression.split(' ');
+  
+    for (const token of tokens) {
+      if (!isNaN(parseFloat(token))) {
+        // If the token is a number, push it to the stack
+        stack.push(parseFloat(token));
+      } else {
+        // If the token is an operator, perform the operation
+        const result = performOperation(token);
```
appgurueu, line 50: Seeing how this is used, you may want `performOperation` to return a bool. I think this would be cleaner and less error-prone than returning null or undefined (which are both falsey).

Data-Structures/Stack/EvaluateExpression.js
```diff
@@ -0,0 +1,66 @@
...
+        if (result === null) {
+          return null; // Invalid expression
+        }
+      }
+    }
+  
+    if (stack.length === 1) {
```
appgurueu, line 57: Could use a ternary here.

Data-Structures/Stack/test/EvaluateExpression.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('evaluateExpression', () => {
+  it('should evaluate a valid expression', () => {
+    const expression = '3 4 * 2 / 5 +'; // (3 * 4) / 2 + 5 = 11
+    const result = evaluateExpression(expression);
+    expect(result).toBe(11);
+  });
+
+  it('should handle division by zero', () => {
+    const expression = '3 0 /'; // Division by zero
```
appgurueu, line 11: The comment here isn't really needed.

Data-Structures/Stack/test/EvaluateExpression.test.js
```diff
@@ -0,0 +1,22 @@
...
+    const result = evaluateExpression(expression);
+    expect(result).toBe(null);
+  });
+
+  it('should handle an invalid expression', () => {
```
appgurueu, line 16: Same here.

Data-Structures/Stack/test/EvaluateExpression.test.js
```diff
@@ -0,0 +1,22 @@
+import { evaluateExpression } from '../EvaluateExpression.js';
+
+describe('evaluateExpression', () => {
+  it('should evaluate a valid expression', () => {
+    const expression = '3 4 * 2 / 5 +'; // (3 * 4) / 2 + 5 = 11
+    const result = evaluateExpression(expression);
```
appgurueu, line 6: I would save myself this variable (I would probably save myself both variables, but the `expression` variable makes the comment with the expression in infix notation neater to write).

appgurueu (commented):
Looks good! It should be made clear that this is postfix in the name / docs. That's my only really "major" point; most of my comments are more or less matters of taste / minor things, the code looks correct as-is.

Some more minor points:

* I'm not sure Data-Structures/Stack is the correct folder (we use this folder for stack data structures, not for algorithms using stacks currently), but I'm not sure which folder (if any) would be more suitable. Evaluation of expressions can be seen as "math", but in programming lingo math is usually reserved for something else. "string" also doesn't cut it. I'd be fine if you created a "misc" folder for this. 
* You could improve the test coverage a bit - for example, also test `-`, test too many numbers left on stack, test no number on stack (but I don't consider this necesssary).

Data-Structures/Stack/EvaluateExpression.js
```diff
@@ -0,0 +1,66 @@
...
+      // The final result should be on the stack
+      return stack[0];
+    } else {
+      return null; // Invalid expression
+    }
+  }
+  
+  export { evaluateExpression };
```
appgurueu, line 65: The indentation here looks off.

Data-Structures/Stack/test/EvaluateExpression.test.js
gaurovgiri: Didn't quite catch this. Can you elaborate? Thank you!

appgurueu: Since you only use the variables once, you could inline their expressions. So you could do `expect(evaluateExpression(expression)).toBe(11)`, or if you were to inline expression as well, `expect(evaluateExpression('3 4 * 2 / 5 +')).toBe(11)`, but then the comment `// (3 * 4) / 2 + 5 = 11` wouldn't be as close to the expression anymore.

appgurueu (approved):
Though the path should probably be something else, but I'm not sure what. As said, I'd be fine with adding a Misc folder for this, then this could be `Misc/EvaluatePostfixExpression.js` or similar.

gaurovgiri:
Sorry, I was doing changes locally and resolving the mentioned comments and hadn't pushed. @appgurueu Can you please review the changes. Thank you.

appgurueu (approved):

PR #1446, opened 2023-10-05
feat: StableMatching Algorithm

Faareh-Ahmed:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Backtracking/StableMatching.js
+++ b/Backtracking/StableMatching.js
@@ -1,0 +1,65 @@
+// This is the Stable Matching Gale Shapley Algorithm 
+// Gale Shapley describes the method of Stable One to One matching
+
+function initFreeMen() {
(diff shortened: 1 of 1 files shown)


appgurueu (commented):
Needs to be properly separated into implementation with an `export`ed function (+ JSDoc comment) and tests. There should be no `console.log`. This should not be operating on a global state - rather, it should be made reusable. I also don't see how this belongs into backtracking. (There is no need to create a class, though.)

Code-wise, the implementation looks like it may be correct, but is needlessly inefficient. This algorithm can be implement in O(n²) by making each loop iteration constant time, you just need to be careful with your data structures. This implementation is probably at least O(n³) because of all the linear searches it does though.

PR #1458, opened 2023-10-06
library sort added

yathu1:
library sort algorithm added


appgurueu:
Your tests are failing (also, have you made sure that we do not already have this algorithm, perhaps under a different name? Just seeing the `gap` in the code reminds me of shell sort, for instance).

PR #1468, opened 2023-10-07
feat:  IterativeDeepeningDepthFirstSearch.js a graph searching algorithm 

W41K3R0:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Graphs/IterativeDeepeningDepthFirstSearch.js
+++ b/Graphs/IterativeDeepeningDepthFirstSearch.js
@@ -1,0 +1,78 @@
+class GraphUnweightedUndirected {
+  // Unweighted Undirected Graph class
+  constructor() {
+    this.connections = {}
+  }
+
+  addNode(node) {
+    // Function to add a node to the graph (connection represented by set)
+    this.connections[node] = new Set()
+  }
+
+  addEdge(node1, node2) {
+    // Function to add an edge (adds the node too if they are not present in the graph)
+    if (!(node1 in this.connections)) {
+      this.addNode(node1)
+    }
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
* Please try to merge this with our existing graph implementations rather than adding yet another one.
* When is IDS useful? At first glance this looks like a very inefficient way to implement breadth-first search to me.
* Missing tests and proper docs. Don't include examples as outcommented code, include them as JSDoc `@example`s or via the tests.

W41K3R0:
Please do not compare this with BFS , it is a modified dfs search ,
 IDS is guaranteed to find a solution if one exists, as long as the search space is finite and the branching factor is not extremely high. It's a complete search algorithm, unlike pure depth-first search (DFS), which can get stuck in infinite branches.
Its understandable that you don't wanna marge it , yeah

appgurueu:
> IDS is guaranteed to find a solution if one exists

This guarantee is very weak. For a finite graph, any algorithm that eventually visits all nodes will have this property, and the graphs we're dealing with here are exclusively finite.

> Please do not compare this with BFS , it is a modified dfs search

What I could discern from the code is:

* You check all nodes until a certain depth
* You increase the depth limit by one
* Rinse and repeat until node is found

This is effectively an inefficient breadth-first search, no? With each iteration of the loop, you check another level.

> It's a complete search algorithm, unlike pure depth-first search (DFS), which can get stuck in infinite branches.

How is that relevant here? The in-memory data structures involved here aren't infinite as far as I can tell.

PR #1470, opened 2023-10-07
Sleep Sort Algorithm

sahilsync07:
DESC:

Add Sleep Sort Algorithm


appgurueu:
We already have a PR for this (https://github.com/TheAlgorithms/JavaScript/pull/1434). The way this is written also makes it practically unusable.

PR #1471, opened 2023-10-07
chore: update maths variables in docstrings from Integer to Number

ridge-kimani:

### Describe your change:

- [ ] Documentation change?
Most docstrings use `Integer` in the maths package as the input params instead of Number. `Number` is the accepted param since it directly correlates with the datatype `Number`

### Checklist:


appgurueu (changes requested):
Many of these were overly specific previously and are in fact not limited to integers, but you also applied this change to many places where this isn't the case (ex. - not exhaustive: liouville function, hexagonal numbers, find max recursion `left` & `right` etc.)

ridge-kimani:
Most of those files have the { Integer } param which isn't a valid type. The acceptable valid type is { Number } and that's the purpose of the changes.

appgurueu:
Nitpick: In the JSDoc examples I'm seeing `number`. Are you sure that `Number` is any more "valid" than `Integer`?

`Integer` isn't an "invalid" type per se. Custom types or class-defined types would be written like this. `Integer` is supposed to mean "a number, but whole and usually within some safe integer range", but we haven't defined that anywhere. Ideally files which use this "type" should contain a [`@typedef`](https://jsdoc.app/tags-typedef.html) tag in the docs. We could also define types like this in one file and have other files require it, but this seems to be [somewhat awkward](https://stackoverflow.com/questions/43183450/jsdoc-typedef-in-a-separate-file).

I don't think just changing all occurrences of `Integer` to `number` is a proper solution since it removes valuable information.

ridge-kimani:
In the typedef it's either `Number` or `number`, it's not case-sensitive.

PR #1473, opened 2023-10-07
bug: abs returns 0 on an empty array

ridge-kimani:

### Describe your change:
- Updates an edge case to return invalid type on objects

### Checklist:

### DIFF
--- a/Maths/Abs.js
+++ b/Maths/Abs.js
@@ -12,7 +12,7 @@
 const abs = (num) => {
   const validNumber = +num // converted to number, also can use - Number(num)
 
-  if (Number.isNaN(validNumber) || typeof num !== "number") {
+  if (Number.isNaN(validNumber) || typeof num === 'object') {
     throw new TypeError('Argument is NaN - Not a Number')
   }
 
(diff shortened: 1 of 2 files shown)


appgurueu (approved):
Not opposed to this as it seems to fix an edge case the author hadn't considered.

`abs(null)` will still be `0` due to the coercion of `+`. Types should probably be checked before coercing, otherwise it makes the most sense to accept JS coercion rules.

Frankly I think this `Abs` function is a mess without algorithmic value though. I think it would probably be best to remove it.

ridge-kimani:
abs(null) will still be caught by the object check since `typeof null` is `object`

appgurueu:
> abs(null) will still be caught by the object check since `typeof null` is `object`

That's true, I missed this. As said, I approve this PR, it makes the function be closer to what the interface was supposed to be. But I think in the future, we should either have (1) strict type checks (in this case: only allow a number) or (2) preferably, no overly defensive type checks, as is idiomatic in JavaScript, and simply assuming the user knows what (not) to do.

raklaptudirm (approved):

PR #1477, opened 2023-10-08
Feat: Format duration ms to duration

developerr-ayush:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:


developerr-ayush:
1. hey can you give me hectoberfest label?
2. do you have any resource for understanding forking if you see above then there are lots of commits visible and that commits are from previous merge request although it doesn't affect but how can I sort this?
3. i have tried the rebase thing but yet didn't worked.

Thanks

Conversions/FormatDuration.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Formats a duration in milliseconds into a human-readable string.
+ *
+ * @param {number} ms - The duration in milliseconds (can be negative).
+ * @returns {string} A formatted string representing the duration.
+ */
+const FormatDuration = (ms) => {
+  if (ms < 0) ms = -ms
```
appgurueu, line 8: Why is this useful or reasonable? I'd remove it altogether and just not allow negative durations.

appgurueu (approved):
Code is fine, has tests and docs. I'm not sure whether we can consider this algorithmically interesting enough, though. The string formatting logic here is pretty straightforward.

appgurueu (commented):
Why did you add a `msToDuration` function here?

developerr-ayush:
@appgurueu it is a related function to the same which converts ms to some other string that's why I have added that here in this pr

appgurueu:
I don't really see a point in having two functions around which do similar things.

Still, the main question persists: What does this - algorithmically - teach a reader?

developerr-ayush:
so we have video playforms in that the second function can be used to show the live duration and the other one is kinda hard to tell what will be helpful from that

developerr-ayush:
i have faced this one i have to make an timer for OTP for that I have to take 30 sec, 5 min , 30 min. so in this case I have made this algo for first function which has names in output. second one was a video player to show live duration counter as I was getting data as ms

raklaptudirm:
That is not the point @developerr-ayush. It maybe a useful modular piece of code, but remember that this repository is not supposed to be used as a module. All the code added here should provide some educational value to the reader, and therefore useful but trivial algorithms won't be accepted.

developerr-ayush:
Okay i think its useless then

PR #1479, opened 2023-10-08
cycle.js

ZainabMalik5:
I have added javaScript file in Graphs folder for detecting cycle in undirected graph.Please accept my PR.


appgurueu (changes requested):
* We have multiple existing graph data structures. Please see how you can merge it with those. This may require cleanup / replacement / adaption of existing code.
* This needs tests and proper documentation. The code should also be commented better.
* Remove the outcommented code. This should be in the tests, or in a JSDoc `@example`.

PR #1480, opened 2023-10-08
Kahn's Algorithm

ZainabMalik5:
I have added a javaScript function for detecting cycles in the directed graph which is Kahn's algorithm. Please merge my PR.


appgurueu (changes requested):
See my review for `cycle.js`, essentially the same points apply here. Also, separate the two things. `cycle.js` does not belong here. Per our naming convention, these file names should not be lowercase. The implementation of Kahn's algorithm given here is also likely needlessly inefficient since `queue.shift()` is a linear time operation. Try to use one of our linked queues.

PR #1482, opened 2023-10-08
implemented CycleDetectionII code in LinkedList

Akshay73c:
Hey maintainers, 
I've implemented the code for getting the starting node of cycle in a linked list which i think will be useful for understanding of cycle-type-questions.
I've also added the tests to check the correctness.
It'll be my first open source contribution and also in _Hacktoberfest_. Hoping a success :)

### Changes:

- [ ] Added a code to get the starting node of cycle in LinkedList

### Checklist:
I've read the CONTRIBUTING.md file and also checked the testcases using npm as asked to. 
This pr is all my own work.


appgurueu (commented):
* The code is still messy (mostly formatting-wise, the `lengthCycle` function should be defined before use for better readability, IMO). Please add comments, and give better variable names (i.e. I assume `f` and `s` are `fast` and `slow`; why do you need them when you already have `fast` and `slow`?).
* As for the tests, remove the 3 redundant comments there which just repeat test case names.

Akshay73c:
Dear maintainers,
Thanks for the feedback !!  I appreciate you taking the time to look over the code and provide suggestions.
I've made the following changes as suggested: 
- Added comments explaining the logic
- you're absolutely right that using f and s is redundant when I already have fast and slow pointers
- removed the redundant comments in tests
- In my opinion, for this problem the key focus is detecting the cycle start node, so I wanted to keep that logic upfront in detectCycleNode(). Defining the helper lengthCycle() function below separates it out from the main detection logic. That said, I'm open to rearranging if you feel strongly that defining lengthCycle() first improves the overall structure.

Let me know! And thanks again for the review :)

I wanted to reach out to let you know that while attempting my first contribution to this open source project, I ran into some issues with git push. (That's why so many commits) After doing some research, I was able to resolve the issues and successfully contribute my changes.
I realize this may not have been the most elegant first contribution, and I wanted to apologize in advance if I caused any inconvenience. As a first time contributor, I was still learning the contribution workflow.

Akshay73c:
Hey maintainers, I'd be really glad if you could review this ;)

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * A LinkedList based solution for Detecting a start node of the Cycle in a list.
```
appgurueu, line 2: I'd say "finding" a start node rather than "detecting" one. Also consider something like `@return a node on a cycle in the linked list`.

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+ * https://en.wikipedia.org/wiki/Cycle_detection
+ */
+
+function detectCycleNode(head) {
+  /*
```
appgurueu, line 7: This comment is redundant except for the Leetcode link (which could be added above). Please use the JSDoc `@see` annotation both for the Wiki & Leetcode link.

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+    slow = slow.next
+    if (fast === slow) {
+      // Calculate length of the cycle
+      length = lengthCycle(slow)
+      break
+    }
+  }
+
+  if (length === 0) {
+    // If length = 0, return null
```
appgurueu, line 28: Redundant comment. A more useful comment might be "if there is no cycle, return `null`".

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+  while (fast !== slow) {
+    fast = fast.next
+    slow = slow.next
+  }
+
+  // return the meeting node
+  return slow
+}
+
+function lengthCycle(head) {
```
appgurueu, line 51: This function seems overcomplicated to me. This function is only called if there is a cycle, yet there is effectively dead/unnecessary code to handle the case that there is no cycle. Also, why do you need another "tortoise-and-hare" here? If `head` already is a node on the cycle, can't you just determine the length of the cycle by following pointers until you're back to `head`? (Effectively your inner loop.)

If I'm missing something and this additional work is necessary, there should be comments justifying / explaining its purpose. It is not self-evident.

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+      break
+    }
+  }
+
+  if (length === 0) {
+    // If length = 0, return null
+    return null
+  }
+
+  //Point the nodes at head
```
appgurueu, line 32: Redundant comment. (Also, could be shortened to `fast = slow = head` in principle.)

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+  fast = head
+  slow = head
+  // Move slow pointer ahead by 'length' of cycle times
```
appgurueu, line 35: The purpose of this is unclear to me. If `length` is in nodes, then `fast = slow = head` after this and this is effectively dead code. If `length` is in "edges" / "links" (one fewer than nodes), then you will get the predecessor of `head` in the cycle, but what for?

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+  while (length > 0) {
+    slow = slow.next
+    length--
+  }
+
+  // Now move both pointers until they meet - this will be the start of cycle
```
appgurueu, line 41: Why will it be the "start" of the cycle?

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,77 @@
...
+  while (fast !== slow) {
+    fast = fast.next
```
appgurueu, line 43: This would have to be `fast.next.next`. If this is not failing the tests in its current form, it means `fast == slow`, and all this code is practically dead.

appgurueu (changes requested):
Of course! This seems to work, but there is definitely dead / unnecessary code, and why the code is doing what it is doing is not exactly clear. Also some minor stylistic remarks.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: Yeah, I've updated the comment to reflect that we are finding the starting node, rather than detecting it. Also added a JSDoc `@returns` comment specifying that the function returns a node on a cycle in the linked list for clarity.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: Yup, your suggested comment would clarify better about the null return value.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: Yes. `fast = slow = head` would be better. I have made this change as well.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: Here, the 'length' refers to the number of nodes to move forward the 'slow' pointer in the linked list. (while keeping the 'fast' pointer at head only ). Then we'll move both pointers one by one to get the starting node of cycle. I have attached an image to help illustrate this point and my thought process.

Akshay73c: Thought Process: 
CycleNode1

Akshay73c: CycleNode2

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: To help illustrate my thought process, please see the attached image above. Hopefully, it'll make things more clear. Let me know if anything is fuzzy.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: I appreciate you pointing out the potential for misinterpretation. Here, the `fast` pointer is kept stationery at the head only and `slow` is moved ahead first in the list. And then we move both one by one to make them meet at the starting node. Please see the attached images above for my justification.

Data-Structures/Linked-List/CycleDetectionII.js
Akshay73c: Thank you for pointing out the potential overcomplication. Though it appears redundant, the inner loop and both pointers are crucial for calculating the cycle length accurately. Even if the head is already on the cycle, we need both the pointers to actually go inside the cycle to determine its length. The condition you mentioned isn't meant to be explicitly checked. It automatically falls into place as both pointers move within the cycle. Also to make the 'length' function more universal or like isolated, I've kept it independent of other functions and reusable.

Akshay73c:
Dear maintainers,
Thanks for reviewing again! I appreciate you taking this much time to provide suggestions for a beginner like me. 
I've made few minor changes to the code and also explained my thought process and addressed to your valuable reviews. Please let me know if anything is required. 
And am really sorry for the delays as i was occupied with campus interviews and exams. (am looking for my first internship)

Akshay73c:
Please let me know if there's anything more that needs to be addressed. Fingers crossed for the merge ;)

Data-Structures/Linked-List/CycleDetectionII.js
appgurueu: > Though it appears redundant, the inner loop and both pointers are crucial for calculating the cycle length accurately.

I never challenged the necessity of the inner loop, quite the opposite.

> Even if the head is already on the cycle, we need both the pointers to actually go inside the cycle to determine its length.

How so? As said, just the inner loop should work.

> Also to make the 'length' function more universal or like isolated, I've kept it independent of other functions and reusable.

This is pointless given that it's not exported.

Even if it was exported (and tested), I don't think duplicating code and responsibilities is a good idea. Finding a node on a cycle and finding the length of the cycle are two different things. There is no need to duplicate the "finding a node on a cycle" (tortoise and hare) logic.

To make sure that my logic is correct, I simplified the `lengthCycle` function to

```javascript
// head is a node on a cycle
function lengthCycle(head) {
  // How long until we visit head again?
  let cur = head;
  let len = 0;
  do {
  	cur = cur.next;
  	len++;
  } while (cur != head);
  return len;
}
```

and as expected, the tests are still passing.

Data-Structures/Linked-List/CycleDetectionII.js
```diff
@@ -0,0 +1,71 @@
...
+      break
+    }
+  }
+
+  if (length === 0) {
+    // If there is no cycle, return null.
+    return null
+  }
+
+  fast = slow = head
```
appgurueu, line 28: This is what confused me. Please don't reuse these variables for something entirely different...

Given that the pointers move at the same speed, but one is "ahead" of the other (until they meet), maybe call these "ahead" and "behind"? The names "slow" and "fast" are misleading.

Data-Structures/Linked-List/CycleDetectionII.js
appgurueu: Thanks for the effort, I understand it now. Maybe the comments could do a better job conveying this?

Data-Structures/Linked-List/CycleDetectionII.js
appgurueu: Yeah, I see why now. Again, see my point about the confusing naming of these variables.

appgurueu (dismissed):

appgurueu (approved):

Akshay73c:
> Sorry for the long delay. I simplified the code a bit and renamed some variables. I hope you're fine with this, otherwise feel free to complain :)
> 
> LGTM!

LGTM too! Thanks for simplifying things, and no complaints here :))

raklaptudirm (approved):

PR #1483, opened 2023-10-09
feat:BidirectionalSearch Algorithm

AmbrishRamachandiran:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Graphs/BidirectionalSearch.js
+++ b/Graphs/BidirectionalSearch.js
@@ -1,8 +1,8 @@
 /*
 Bidirectional search is a graph search algorithm that finds a shortest path from an initial vertex to a goal vertex in a directed graph. 
 It runs two simultaneous searches: one forward from the initial state, and one backward from the goal, stopping when the two meet. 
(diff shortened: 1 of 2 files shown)


Graphs/BidirectionalSearch.js
```diff
@@ -0,0 +1,68 @@
+/*
+Bidirectional search is a graph search algorithm that finds a shortest path from an initial vertex to a goal vertex in a directed graph. 
+It runs two simultaneous searches: one forward from the initial state, and one backward from the goal, stopping when the two meet. 
+The reason for this approach is that in many cases it is faster: for instance, in a simplified model of search problem complexity in which both searches expand a tree with branching factor b, and the distance from start to goal is d, each of the two searches has complexity O(bd/2) (in Big O notation),
```
appgurueu, line 4: The big-O here doesn't make sense. You probably mean O(b^(d/2)) and O(b^d).

appgurueu (changes requested):
1. Don't create yet another adjacency list representation. Either use / adapt an existing one, or just expect your input to be in the proper format (e.g. array of arrays).
2. In accordance with 1, `bidirectionalSearch` should just be a function or a method on an adjacency-list based `Graph` class (which we already have, look into the existing graph data structures).
3. You're using `shift` to pop the first element from the queues, which has linear time complexity. Please use our linked queues or similar to achieve constant time complexity.

PR #1513, opened 2023-10-11
Compression: Huffman Coding

aladin002dz:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Compression/test/HuffmanArray.test.js
+++ b/Compression/test/HuffmanArray.test.js
@@ -25,11 +25,6 @@ describe('Huffman Coding', () => {
   it('should build Huffman codes correctly', () => {
     const codes = buildHuffmanCodes(root)
 
-    console.log(codes['t'])
-    console.log(codes['h'])
-    console.log(codes['i'])
-    console.log(codes['s'])
-
     expect(codes['t']).toEqual('01101')
     expect(codes['h']).toEqual('0111')
     expect(codes['i']).toEqual('0100')


Compression/Huffman.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {Object} freqTable - The frequency table to use for building the tree.
+ * @returns {HuffmanNode} - The root node of the Huffman tree.
+ */
+function buildHuffmanTree(freqTable) {
+  const nodes = Object.keys(freqTable).map(
+    (char) => new HuffmanNode(char, freqTable[char])
+  )
+
+  while (nodes.length > 1) {
+    nodes.sort((a, b) => a.freq - b.freq)
```
appgurueu, line 57: Why do you sort in every iteration?

Compression/Huffman.js
```diff
@@ -0,0 +1,146 @@
...
+    const left = nodes.shift()
```
appgurueu, line 58: Unnecessarily linear time: If you sort in reverse order, you can use a constant time `pop` here instead. Same for the line below.

Compression/Huffman.js
```diff
@@ -0,0 +1,146 @@
...
+    if (currentNode.char) {
+      decodedData += currentNode.char
+      currentNode = root
+    }
+  }
+
+  return decodedData
+}
+
+// Example usage
```
appgurueu, line 130: This does not belong here. It should be in the tests (where it already is) / JSDoc examples (where some methods could still use some of it).

appgurueu (commented):
The other main point is string handling, which pretty consistently uses repeated concatenation here (e.g. for building prefixes (that one is mostly a non-concern though, since even with slow string concat, it'll be linearithmic time), for encoding / decoding). At least the encoding could be rewritten to use an array without making it any less readable. 

Arguably all of this is fine in practice, as JS engines optimize string concatenation. I think it needs some comments alluding to this, though (I'm thinking something along the lines of "the string concatenation in a loop here is fine despite strings being immutable, as modern JS engines optimize it to be amortized constant time by delaying it (using ropes) and similar techniques").

I'm sorry about the unrelated changes due to formatting, that's due to a mistake on our part (see https://github.com/TheAlgorithms/JavaScript/pull/1515).

Keep up the good work!

Compression/Huffman.js
aladin002dz: because with the push on a new node, the order may change.

appgurueu: Wouldn't it be appropriate to use a max heap here (our existing implementations should work, you just need to import and use them), given that you always extract the most frequent ones?

aladin002dz: I think it would add a lot more complexity, wouldn't it?

appgurueu: I think the API of our heap should be pretty straightforward to use. A heap is pretty much *the* data structure for this use case; it would significantly help the time complexity.

aladin002dz: I'll try to implement it myself, then I'll need your valuable feedback and guidance.

appgurueu (commented):
Why have you duplicated the code? I think there should only be a heap-based implementation; it is strictly better (more efficient) than the array-based one and of similar code complexity.

aladin002dz:
> Why have you duplicated the code? I think there should only be a heap-based implementation; it is strictly better (more efficient) than the array-based one and of similar code complexity.
Duplicate removed.

appgurueu:
Adding the hacktoberfest-accepted label since we're close to the end of October, and this PR is pretty close to being accepted and merged later on.

PR #1519, opened 2023-10-12
feat: Added MD5 hashing algorithm

ManpreetSingh2004:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Hashes/MD5.js
+++ b/Hashes/MD5.js
@@ -1,0 +1,218 @@
+//= ===============================================================
+// MD5.js
+//
+// Module that replicates the MD5 Cryptographic Hash
+// function in Javascript.
+//= ===============================================================
+
+// main variables
+const CHAR_SIZE = 8
+
+const S = [
+  7, 12, 17, 22, 7, 12, 17, 22, 7, 12, 17, 22, 7, 12, 17, 22, 5, 9, 14, 20, 5,
+  9, 14, 20, 5, 9, 14, 20, 5, 9, 14, 20, 4, 11, 16, 23, 4, 11, 16, 23, 4, 11,
+  16, 23, 4, 11, 16, 23, 6, 10, 15, 21, 6, 10, 15, 21, 6, 10, 15, 21, 6, 10, 15,
+  21
+]
+
+const K = [
+  0xd76aa478, 0xe8c7b756, 0x242070db, 0xc1bdceee, 0xf57c0faf, 0x4787c62a,
+  0xa8304613, 0xfd469501, 0x698098d8, 0x8b44f7af, 0xffff5bb1, 0x895cd7be,
+  0x6b901122, 0xfd987193, 0xa679438e, 0x49b40821, 0xf61e2562, 0xc040b340,
+  0x265e5a51, 0xe9b6c7aa, 0xd62f105d, 0x02441453, 0xd8a1e681, 0xe7d3fbc8,
+  0x21e1cde6, 0xc33707d6, 0xf4d50d87, 0x455a14ed, 0xa9e3e905, 0xfcefa3f8,
+  0x676f02d9, 0x8d2a4c8a, 0xfffa3942, 0x8771f681, 0x6d9d6122, 0xfde5380c,
+  0xa4beea44, 0x4bdecfa9, 0xf6bb4b60, 0xbebfbc70, 0x289b7ec6, 0xeaa127fa,
+  0xd4ef3085, 0x04881d05, 0xd9d4d039, 0xe6db99e5, 0x1fa27cf8, 0xc4ac5665,
+  0xf4292244, 0x432aff97, 0xab9423a7, 0xfc93a039, 0x655b59c3, 0x8f0ccc92,
+  0xffeff47d, 0x85845dd1, 0x6fa87e4f, 0xfe2ce6e0, 0xa3014314, 0x4e0811a1,
+  0xf7537e82, 0xbd3af235, 0x2ad7d2bb, 0xeb86d391
+]
+
+/**
+ * Converts the string to little endian in groups of 8
+ *
+ * @param {string} str - 32 character string representation (binary)
+ * @returns {string} - little endian string
+ *
+ * @example
+ *      toLittleEndian("11111111111110100011100101000010"); // "01000010001110011111101011111111"
+ */
+function toLittleEndian(str) {
+  return (
+    str.slice(24, 32) + str.slice(16, 24) + str.slice(8, 16) + str.slice(0, 8)
+  )
+}
+
+/**
+ * Converts a positive integer to a hexadecimal string representation
+ *
+ * @param {number} num - A 32 bit positive integer
+ * @returns {string} - hex representation of the number
+ *
+ * @example
+ *      toHex(255); // "ff000000"
+ */
+function toHex(num) {
+  let str = pad(num.toString(16), 8)
+  return str.slice(6, 8) + str.slice(4, 6) + str.slice(2, 4) + str.slice(0, 2)
+}
+
+/**
+ * Adds padding to binary/hex string representation
+ *
+ * @param {string} str - string representation (binary/hex)
+ * @param {int} bits - total number of bits wanted
+ * @return {string} - string representation padding with empty (0) bits
+ *
+ * @example
+ *      pad("10011", 8); // "00010011"
+ */
+function pad(str, bits) {
+  let res = str
+  while (res.length % bits !== 0) {
+    res = '0' + res
+  }
+  return res
+}
+
+/**
+ * Separates string into chunks of the same size
+ *
+ * @param {string} str - string to separate into chunks
+ * @param {int} size - number of characters wanted in each chunk
+ * @return {array} - array of original string split into chunks
+ *
+ * @example
+ *      chunkify("this is a test", 2)
+ */
+function chunkify(str, size) {
+  const chunks = []
+  for (let i = 0; i < str.length; i += size) {
+    chunks.push(str.slice(i, i + size))
+  }
+  return chunks
+}
+
+/**
+ * Rotates the bits to the left
+ *
+ * @param {number} bits - 32 bit number
+ * @param {int} turns - number of rotations to make
+ * @return {string} - number after bits rotation
+ *
+ * @example
+ *      rotateLeft("1011", 3); // "1101"
+ */
+function rotateLeft(bits, turns) {
+  return (bits << turns) | (bits >>> (32 - turns))
+}
+
+/**
+ * Pre-processes message to feed the algorithm loop
+ *
+ * @param {string} message - message to pre-process
+ * @return {string} - processed message
+ */
+function preProcess(message) {
+  // convert message to binary representation padded to
+  // 8 bits, and add 1
+  let m =
+    message
+      .split('')
+      .map((e) => e.charCodeAt(0))
+      .map((e) => e.toString(2))
+      .map((e) => pad(e, 8))
+      .join('') + '1'
+
+  // extend message by adding empty bits (0)
+  while (m.length % 512 !== 448) {
+    m += '0'
+  }
+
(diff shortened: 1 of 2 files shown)


Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
+//= ===============================================================
```
appgurueu, line 1: This line and the following is just spam.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+// MD5.js
+//
+// Module that replicates the MD5 Cryptographic Hash
```
appgurueu, line 4: This is the only useful comment in this block.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+  0xd4ef3085, 0x04881d05, 0xd9d4d039, 0xe6db99e5, 0x1fa27cf8, 0xc4ac5665,
+  0xf4292244, 0x432aff97, 0xab9423a7, 0xfc93a039, 0x655b59c3, 0x8f0ccc92,
+  0xffeff47d, 0x85845dd1, 0x6fa87e4f, 0xfe2ce6e0, 0xa3014314, 0x4e0811a1,
+  0xf7537e82, 0xbd3af235, 0x2ad7d2bb, 0xeb86d391
+]
+
+/**
+ * Converts the string to little endian in groups of 8
+ *
+ * @param {string} str - 32 character string representation (binary)
```
appgurueu, line 35: That's a pretty inefficient representation to work with. Why not use a single number instead? "Slicing" would then be done much more efficiently using bitops.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+  return (
+    str.slice(24, 32) + str.slice(16, 24) + str.slice(8, 16) + str.slice(0, 8)
+  )
+}
+
+/**
+ * Converts a positive integer to a hexadecimal string representation
+ *
+ * @param {number} num - A 32 bit positive integer
+ * @returns {string} - hex representation of the number
```
appgurueu, line 51: Yet another representation. Why do you need this?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+ * @param {int} bits - total number of bits wanted
+ * @return {string} - string representation padding with empty (0) bits
+ *
+ * @example
+ *      pad("10011", 8); // "00010011"
+ */
+function pad(str, bits) {
+  let res = str
+  while (res.length % bits !== 0) {
+    res = '0' + res
```
appgurueu, line 74: Repeated string concatenation is a theoretically not necessarily efficient (but practically efficient) way to build strings in JavaScript; I don't think it's necessarily good practice to use it here. Just something like `'0'.repeat(res.length % bits) + str` should be cleaner.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+  for (let i = 0; i < str.length; i += size) {
+    chunks.push(str.slice(i, i + size))
+  }
+  return chunks
+}
+
+/**
+ * Rotates the bits to the left
+ *
+ * @param {number} bits - 32 bit number
```
appgurueu, line 100: Here you use 32-bit numbers now.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+  let m =
+    message
+      .split('')
+      .map((e) => e.charCodeAt(0))
+      .map((e) => e.toString(2))
+      .map((e) => pad(e, 8))
+      .join('') + '1'
+
+  // extend message by adding empty bits (0)
+  while (m.length % 512 !== 448) {
```
appgurueu, line 129: This should've used the pad function, no?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+ */
+function preProcess(message) {
+  // convert message to binary representation padded to
+  // 8 bits, and add 1
+  let m =
+    message
+      .split('')
+      .map((e) => e.charCodeAt(0))
+      .map((e) => e.toString(2))
+      .map((e) => pad(e, 8))
```
appgurueu, line 125: This is fragile. It won't work as expected for characters that use more than 8 bits in UTF-16. Strings are ultimately the wrong type to use here.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+  m = chunkify(m, 32)
+    .map((e) => toLittleEndian(e))
+    .join('')
+
+  // length of message in binary, padded, and extended
+  // to a 64 bit representation
+  let ml = (message.length * CHAR_SIZE).toString(2)
+  ml = pad(ml, 64)
+
+  return m + ml.slice(32) + ml.slice(0, 32)
```
appgurueu, line 142: Again some endianness reversal? Why?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+}
+
+/**
+ * Hashes message using MD5 Cryptographic Hash Function
+ *
+ * @see
+ * For more info: https://en.wikipedia.org/wiki/MD5
+ * 
+ * @param {string} message - message to hash
```
appgurueu, line 151: Both the message and the return value should be byte arrays (`Uint8Array`). String simply isn't the proper type here; JS strings aren't byte strings.

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+    const abcd = Uint32Array.from([
+      hashes[a0],
+      hashes[b0],
+      hashes[c0],
+      hashes[d0]
+    ])
+
+    for (let i = 0; i < 64; i++) {
+      const [F, g] = [0, 1]
+      const fg = Uint32Array.from([0, 0])
```
appgurueu, line 182: An Uint32Array for two variables? Why?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+      }
+
+      fg[F] = fg[F] + abcd[A] + K[i] + words[fg[g]]
+      abcd[A] = abcd[D]
+      abcd[D] = abcd[C]
+      abcd[C] = abcd[B]
+      abcd[B] = abcd[B] + rotateLeft(fg[F], S[i])
+    }
+
+    // add values for this chunk to main hash variables (unsigned)
```
appgurueu, line 205: 1. Use `+=`
2. Shouldn't this be something like wrapping addition? What happens if these grow beyond 32 bits?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * @see
+ * For more info: https://en.wikipedia.org/wiki/MD5
+ * 
+ * @param {string} message - message to hash
+ * @return {string} - message digest (hash value)
+ */
+function MD5(message) {
+  // Initialize variables:
+  const [a0, b0, c0, d0] = [0, 1, 2, 3]
```
appgurueu, line 156: Why are the variables for these indices useful? They only seem to make the code harder to read. If you want to name these for some reason, why don't you make the 4 hashes variables which you can name instead?

Hashes/MD5.js
```diff
@@ -0,0 +1,221 @@
...
+    // add values for this chunk to main hash variables (unsigned)
+    hashes[a0] = hashes[a0] + abcd[A]
+    hashes[b0] = hashes[b0] + abcd[B]
+    hashes[c0] = hashes[c0] + abcd[C]
+    hashes[d0] = hashes[d0] + abcd[D]
+  })
+
+  // combine hash values of main hash variables and return
+  const HH = [hashes[a0], hashes[b0], hashes[c0], hashes[d0]]
+    .map((e) => toHex(e))
```
appgurueu, line 214: This has no place here if this function is changed to produce a `Uint8Array`.

appgurueu (changes requested):
The code is pretty messy because it juggles various representations of bytes that aren't really suitable (strings of bits encoded as ASCII 0 / 1 etc.)
It should be changed to just use [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) wherever byte arrays are needed, and simply numbers wherever 4 byte "words" are needed. The string representations (and the code for dealing with them) should be gotten rid of entirely.

Also needs tests.

ManpreetSingh2004:
Hi @appgurueu,
Thanks for the feedback!
Actually I'm quite new to the js world. I looked at the SHA implementations in the nearby files and assumed that it's the right way to do it. I'll try my best to fix the issues and make the code cleaner.

appgurueu:
> [...] I looked at the SHA implementations in the nearby files and assumed that it's the right way to do it. [...]

Ah, sorry about that. We have much legacy code lying around that needs to be refactored and definitely isn't a good example. We should make it more clear in the contribution guidelines that much of the legacy codebase shouldn't be taken as a good example.

Hashes/MD5.js
ManpreetSingh2004: Spam has been removed.

Hashes/MD5.js
ManpreetSingh2004: All string-related functions have been removed.

Hashes/MD5.js
ManpreetSingh2004: This has been retained. Now we use 32 bit numbers almost everywhere.

Hashes/MD5.js
ManpreetSingh2004: It is a little bit different. But now we use the `PadEnd` function here.

Hashes/MD5.js
ManpreetSingh2004: This is a required preprocessing step. See [The Algorithms Python implementation](https://github.com/TheAlgorithms/Python/blob/5645084dcd5cf398caefa40641ac99144a40e572/hashes/md5.py#L129C85-L129C85)

Hashes/MD5.js
ManpreetSingh2004: Uint32 Arrays were used to avoid manual wrapping addition, and (type casting?) using `num >>> 0`.
By using typed arrays we ensure that the numbers stored are 32 bit unsigned integers. I opted for typed arrays as an alternative to using the >>> operator due to my lack of confidence in its usage.

ManpreetSingh2004:
Hi @appgurueu,
I believe I have implemented all of the requested changes. Please let me know if any additional adjustments are needed.

Hashes/tests/MD5.test.js
```diff
@@ -0,0 +1,37 @@
+import { MD5 } from '../MD5'
+
+describe('Testing MD5 function', () => {
+  it('should return the correct hash for "The quick brown fox jumps over the lazy dog"', () => {
+    const input = new TextEncoder().encode(
+      'The quick brown fox jumps over the lazy dog'
+    )
+
+    const hash = Array.from(MD5(input), (byte) =>
```
appgurueu, line 9: Why not introduce a function `hexMD5` that does this? Duplicating this three times doesn't seem very viable.

Hashes/tests/MD5.test.js
```diff
@@ -0,0 +1,37 @@
...
+      byte.toString(16).padStart(2, '0')
+    ).join('')
+
+    expect(hash).toBe('9e107d9d372bb6826bd81d3542a419d6')
+  })
+
+  it('should return the correct hash for "The quick brown fox jumps over the lazy dog."', () => {
```
appgurueu, line 16: Took me a moment to notice the period. Why not use a more distinct example?

appgurueu:
Adding the hacktoberfest-accepted label in case we don't get this merged before October ends; the PR is almost there!

appgurueu (approved):

raklaptudirm (approved):

ManpreetSingh2004:
Thanks!

appgurueu:
Thank you for your contribution!

PR #1521, opened 2023-10-13
Create knapsack.js

100mya:
created knapsack.js - solving knapsack problem using backtracking algorithm


appgurueu (changes requested):
* The class is unnecessary. Refactor this to be a single function which gets values, weights, and capacity as parameters, and returns the max. knapsack value.
* Needs tests and a JSDoc comment.

PR #1522, opened 2023-10-13
Create minmax.js

100mya:
javascript code for minmax problem using backtracking

### DIFF
--- a/Backtracking/minmax.js
+++ b/Backtracking/minmax.js
@@ -1,0 +1,32 @@
+function minimax(depth, nodeIndex, isMax, scores, height) {
+    if (depth < 0) {
+        throw new Error("Depth cannot be less than 0");
+    }
+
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
* Needs a doc comment, esp. given the odd interface (expects a binary tree in array format + height; height is technically redundant).
* Needs proper tests rather than an example at the end of the file.
* Formatting seems off (we use prettier now, there should be no semicolons).
* Why is this in "backtracking"? It's just recursive as far as I can see.
* Why limit minimax to full binary trees stored in an array format?
* The code in the two branches of `if (isMax)` - `else` is unnecessarily duplicated. It should instead just do recursive calls of `minimax(depth + 1, <child index>, !isMax, scores, height)`. You can do `return (isMax ? Math.max : Math.min)(leftValue, rightValue)` then.

PR #1523, opened 2023-10-13
Create hamiltonian_cycle.js

100mya:
javascript code for hamiltonian cycle using backtracking

### DIFF
--- a/Backtracking/hamiltonian_cycle.js
+++ b/Backtracking/hamiltonian_cycle.js
@@ -1,0 +1,77 @@
+/**
+ * Determines whether it is possible to add a next vertex into the path
+ * by validating two conditions:
+ * 1. There should be a path between the current and next vertices.
+ * 2. The next vertex should not be already in the path.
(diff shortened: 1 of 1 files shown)


Backtracking/hamiltonian_cycle.js
```diff
@@ -0,0 +1,77 @@
...
+ * @returns {number[]} - An array of vertices representing the Hamiltonian cycle.
+ */
+function hamiltonCycle(graph, startIndex = 0) {
+    const path = new Array(graph.length + 1).fill(-1);
+    path[0] = path[graph.length] = startIndex; // Initialize path
+
+    return utilHamiltonCycle(graph, path, 1) ? path : [];
+}
+
+// Example usage:
```
appgurueu, line 63: This should be a JSDoc `@example` comment or a test case.

appgurueu (changes requested):
Needs proper tests for both cases (cycle exists / cycle doesn't exist).

PR #1524, opened 2023-10-13
Create all_possible_subsequences.js

100mya:
javascript code for all possible subsequences using backtracking

### DIFF
--- a/Backtracking/all_possible_subsequences.js
+++ b/Backtracking/all_possible_subsequences.js
@@ -1,0 +1,40 @@
+/**
+ * Generates all possible subsequences of the given sequence using backtracking.
+ * Prints each subsequence as it is generated.
+ *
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
Side effects (`console.log`) are a poor API for this. Ideally this should be a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) for much better reusability and testability. A callback or pushing to an array may also be acceptable (though I'd prefer the generator).

Needs proper tests rather than an example in the file.

Also seems to be partially redundant with [`Recursive/SubsequenceRecursive.js`](https://github.com/TheAlgorithms/JavaScript/blob/master/Recursive/SubsequenceRecursive.js), which solves a more specific problem (strings, only nonempty subsequences). Ideally `SubsequenceRecursive.js` should be rewritten in terms of this.

PR #1525, opened 2023-10-13
Create bi_directional_dijkstra.js

100mya:
bi directional dijkstra code in javascript using backtracking


appgurueu (changes requested):
* Interface-wise, shouldn't this take a single graph (the "forward" representation), and generate the "backward" representation from that?
* Implementation-wise, you manage parent arrays, but you do not seem to use them for finding paths; you just return the distance. Can the parent arrays safely be removed, or are you planning to add code to find paths?
* The example should be moved to a proper test.

PR #1530, opened 2023-10-14
Update AllCombinationsOfSizeK.js

imrahulkb:

### Describe your change:

This PR modifies a current algorithm for  **AllCombinationsOfSizeK**. Four parameters made up the original **findCombinations()** method, and they were duplicated for each stack call. Since the modifications are kept in the instance variables, the modified **findCombinations()** method does not require any parameters. In this case, there will be **far fewer** local variables that need to be initialized overall on each recursive call.
There was also a for loop in the original technique. The updates function is **entirely recursive** and lacks a for loop.

### Checklist:

### DIFF
--- a/Backtracking/AllCombinationsOfSizeK.js
+++ b/Backtracking/AllCombinationsOfSizeK.js
@@ -27,18 +27,21 @@ class Combinations {
     this.k = k
     this.current = [] // will be used for storing current combination
     this.combinations = []
+    this.i = 1
   }
 
-  findCombinations(high = this.n, total = this.k, low = 1) {
-    if (total === 0) {
-      this.combinations.push([...this.current])
-      return this.combinations
-    }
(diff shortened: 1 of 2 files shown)


appgurueu (changes requested):
Looks good, however formatting seems broken. If you're already refactoring this, could you get rid of the class entirely? This should be a single function taking `n` and `k` as parameters instead. Ideally the API should also be a generator rather than producing a list.

appgurueu (approved):

appgurueu (changes requested):
You'll need to fix the tests, though. They currently still expect a class.

imrahulkb:
I am not able to get that why there is code style error in this pull request

appgurueu (approved):

raklaptudirm (approved):

PR #1534, opened 2023-10-14
Create AnyBase

imrahulkb:

### Describe your change:

This is a method for converting numbers from one base to another when both bases are in the range 2 to 10.
Because the procedure employed here is more broad, it requires both bases when generating the object. The convert function is used to convert numbers from one base to another.
Internally, there is a micro convert function that will work intermediately because the core algorithm requires one of the bases to be 10, thus the number will be converted to base 10 first and then to the appropriate base.
Unwanted interference can be avoided by using if else, as in if one of the bases is 10.

### Checklist:

### DIFF
--- a/Conversions/AnyBase
+++ b/Conversions/AnyBase
@@ -1,0 +1,22 @@
+class AnyBase {
+  constructor(base1, base2) {
+    this.base1 = base1;
+    this.base2 = base2;
+  }
+
+  convert(num) {
+    const microConvert = (base1, base2, n) => {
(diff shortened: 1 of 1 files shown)


imrahulkb:
I am facing errors due to style of the test file i have created, kindly guide me with this problem

raklaptudirm:
Since the tests passed I assume you have solved the style issue?

imrahulkb:
Yes, I have solved the issue in the latest commit.

PR #1548, opened 2023-10-16
Update Remove Unused Code and Tidy The Code

riancintiyo:
### Describe your change:

- [ ] Remove Unused code
- [ ] Tidy the code


Data-Structures/Graph/Graph2.js
```diff
@@ -13,9 +13,6 @@ class Graph {
   // addEdge(v, w)
```
appgurueu, line 13: shouldn't all of these be deleted, since they are implemented? it doesn't seem to make much sense to only delete the two functions that *aren't* implemented from this list. either remove this list completely or remove just the functions that are already implemented from it.

PR #1555, opened 2023-10-20
Added Time Cache

aryan0078:

### Describe your change:

I have Added TimeCache Which can cache items in Map for provided time period and can destroy itself after that particular time it's efficient way of storing data if you have large amount of values to be stored in cache

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Cache/TimeCache.js
+++ b/Cache/TimeCache.js
@@ -1,0 +1,40 @@
+/* The TimeCache class is a JavaScript class that allows you to store key-value pairs in a cache with
+an optional expiration time. */
+class TimeCache {
+    constructor(expirationTime) {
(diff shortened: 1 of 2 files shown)


aryan0078:
@raklaptudirm 
@appgurueu 
Please review

appgurueu (changes requested):
Sorry, but I'm afraid a wrapper around a map - while perhaps useful in practice - is not really a good fit for this repo; its implementation here is questionable (keys are only removed when you access them - this may lead to memory leaks and largely defeats the purpose of letting keys expire at all), and it is not very reusable - in most algorithms, wall time is not and should not be a factor.

PR #1580, opened 2023-10-26
Create SwapWithoutTemp

imshreya2000:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Bit-Manipulation/SwapWithoutTemp
+++ b/Bit-Manipulation/SwapWithoutTemp
@@ -1,0 +1,14 @@
+function swapWithoutTemp(a, b) {
+    a = a ^ b;   // Step 1: XOR a and b, store result in a
+    b = a ^ b;   // Step 2: XOR new a (a ^ b) and original b, store result in b (now b = a)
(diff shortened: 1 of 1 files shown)


appgurueu (changes requested):
This boils down to

```javascript
a ^= b
b ^= a // a ^ b ^ b = a
a ^= b // a^b^a = b
```

This is a nifty trick to know when for some reason, you need to avoid temporary variables / registers at all costs. But even then it's not necessarily useful; on x86 for example, alternative instructions like `XCHG` tend to be faster.

In JavaScript, it's not really useful; it will probably be slower than a normal swap. Also, there are no proper macros or similar; to make this function usable, you'd have to return `[a, b]`, at which point you could just return `[b, a]` to swap a and b.

Besides, this is missing documentation and tests (which would fail, since JS is pass by value). Leaning towards rejection. @raklaptudirm what do you think?

raklaptudirm:
The xor trick might be good content educationally speaking. That said, this pr is definitely not up to the standards we require here and the API is faulty. I want to keep this pr open in case they can change the code and the API to work well because I would welcome the inclusion of this idea into the repo.

PR #1582, opened 2023-10-26
Implemented Partition Problem, Recursive problem

vedas-dixit:

### Describe your change:

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:


vedas-dixit:
Hello @appgurueu & @raklaptudirm, Can you guys please review "Tug of war-backtracking problem" that I committed recently and leave a comment if any change or modification is required :)

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
+// https://www.geeksforgeeks.org/tug-of-war/
```
appgurueu, line 1: This should be an `@see` annotation in the doc comment.

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+
+// Given an array of n integers, we need to divide them into two subsets such that
```
appgurueu, line 3: This also belongs in the doc comment - it describes what the function does.

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+// the difference between their sums is minimized.
+
+/**
+ * @param {number[]} arr - The array of integers.
+ * @returns {Array.<number[]>} Two subsets with minimized sum difference.
+ */
+
+
+function tugOfWar(arr) {
+    const n = arr.length;
```
appgurueu, line 13: `n` is only used twice, and one usage can be eliminated => this variable should be removed, it doesn't help readability.

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+    let totalSum = arr.reduce((a, b) => a + b, 0);
+    let minimumDiff = Number.MAX_VALUE;
+    let result = [];
```
appgurueu, line 16: Why not start out with `result = [arr, []]` (put `arr` into one subset, other subset is empty), and set `minimumDiff` to `totalSum`?

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @param {number[]} arr - The array of integers.
+ * @returns {Array.<number[]>} Two subsets with minimized sum difference.
+ */
+
+
+function tugOfWar(arr) {
+    const n = arr.length;
+    let totalSum = arr.reduce((a, b) => a + b, 0);
```
appgurueu, line 14: This should be `const`

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+    let minimumDiff = Number.MAX_VALUE;
+    let result = [];
+
+    function isSafe(index, subset) {
```
appgurueu, line 18: This function is unnecessary. It just makes the code less readable; the `subset` parameter is unused.

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+
+            if (currentDiff < minimumDiff) {
+                minimumDiff = currentDiff;
+                result = [subset, arr.filter(x => !subset.includes(x))];
+            }
+            return;
+        }
+
+        // Include the current element.
+        if (isSafe(index, subset)) {
```
appgurueu, line 36: This check is unnecessary. `index < n` will always hold if you reach this point.

Backtracking/TugOfWar.js
```diff
@@ -0,0 +1,50 @@
...
+            backtrack([...subset, arr[index]], index + 1);
+        }
+
+        // Exclude the current element.
+        if (isSafe(index, subset)) {
```
appgurueu, line 41: Same here.

appgurueu (changes requested):
I think this should be categorized as "Recursive" rather than "Backtracking". One definition of *Backtracking* is:

> Backtracking is a class of algorithms for finding solutions to some computational problems, notably constraint satisfaction problems, that incrementally builds candidates to the solutions, and *abandons a candidate as soon as it determines that the candidate cannot possibly be completed to a valid solution.*

I don't see the emphasized part here. To me, this is just a recursive algorithm which tries all possible ways to partition the set into two subsets, evaluating them and picking the best, solving an *optimization problem*.

vedas-dixit:
I was also thinking of putting it in the "Recursive" Questions but this question is labelled as "Backtracking" so I just followed the trend.
Thanks for the review, I'll make the necessary changes :)

vedas-dixit:
I've made some necessary changes @appgurueu 

0. Changed File Location of "TugOfWar.js" under "Recursive" from "Backtracking"
1. Removed unused variable n for clarity.
2. Initialized result with the entire array in one subset and an empty second subset, setting minimumDiff to totalSum.
3. Changed totalSum to const for immutability.
4. Deleted the redundant isSafe function and its associated checks.

vedas-dixit:
I've made some necessary changes @appgurueu & @raklaptudirm 

Changed File Location of "TugOfWar.js" under "Recursive" from "Backtracking"
Removed unused variable n for clarity.
Initialized result with the entire array in one subset and an empty second subset, setting minimumDiff to totalSum.
Changed totalSum to const for immutability.
Deleted the redundant isSafe function and its associated checks.

appgurueu (changes requested):
Ah, I see the issue now. This does not solve the "tug of war" problem; you are missing the side condition that both sets should have size n/2.

This does solve the "partition" problem (which can be reduced to the 0-1-knapsack problem) however.

I suggest either renaming this to `partition` - and removing the GfG reference - or fixing the implementation s.t. the subsets each have size n/2 (note: if n is odd, the sizes may differ by one).

vedas-dixit:
I appreciate your feedback @appgurueu 
I've made the suggested changes to the code. It's now the "Partition Problem," and I've added error handling for invalid input. You can Check out the updated code and tests :)

appgurueu (approved):
(Note: Calculating the sum of nums in each recursive call is wasteful. But the implementation is correct, and really "efficiently" solving the problem isn't possible anyways (unless P = NP), so I don't mind terribly much. Ideally you should have an outer function, which takes just `nums` as a parameter, and an inner function, which takes just `index` and `target`.)

PR #1585, opened 2023-10-27
fix: CI code style checking

appgurueu:
This PR makes the CI use `--check` rather than `--write` so that CI fails if code isn't formatted properly.


appgurueu:
(note: don't squash, just rebase)

raklaptudirm (approved):

PR #1627, opened 2024-03-01
[DEV] Implementation of Hash Table Data Structure

miladsade96:

### Describe your change:

- [x] Adding a data structure called hash table

### DIFF
--- a/Data-Structures/HashTable/HashTable.js
+++ b/Data-Structures/HashTable/HashTable.js
@@ -1,67 +1,95 @@
 export default class HashTable {
   constructor(limit = 100) {
+    // Initialize the storage and limit variables
     this.storage = new Array(limit)
     this.limit = limit
   }
 
+  // Hash function:
   _hash(key, max) {
+    // Initialize the hash variable to 0
     let hash = 0
+    // Iterate through the key and add the character code at each iteration to the hash variable
     for (let i = 0; i < key.length; i++) hash += key.charCodeAt(i)
+    // Return the hash modulo the max
     return hash % max
   }
 
+  // Insert a key-value pair into the hash table
   set(key, value) {
+    // Hash the key
     const index = this._hash(key, this.limit)
+    // If the index is empty, insert the key-value pair and create a bucket
     if (this.storage[index] === undefined) this.storage[index] = [[key, value]]
     else {
+      //  If the index is not empty, iterate through the bucket (collision handling)
       let inserted = false
       for (let i = 0; i < this.storage[index].length; i++) {
         if (this.storage[index][i][0] === key) {
+          // If the key exists, update the value
           this.storage[index][i][1] = value
           inserted = true
         }
       }
+      // If the key does not exist, insert the key-value pair
       if (inserted === false) this.storage[index].push([key, value])
     }
   }
 
   // Get a value from the hash table
   get(key) {
+    // Hash the key
     const index = this._hash(key, this.limit)
+    // If the index is empty, return undefined
     if (this.storage[index] === undefined) return undefined
     else {
+      // If the index is not empty, iterate through the bucket
       for (let i = 0; i < this.storage[index].length; i++) {
+        // If the key exists, return the value
         if (this.storage[index][i][0] === key) return this.storage[index][i][1]
       }
     }
   }
 
+  // Remove a key-value pair from the hash table
   remove(key) {
+    // Hash the key
     const index = this._hash(key, this.limit)
+    // Check if the bucket exists
     if (this.storage[index]) {
+      // If the key matches the key at the index and there is only one item in the bucket, delete the bucket
       if (
         this.storage[index].length === 1 &&
         this.storage[index][0][0] === key
       ) {
         delete this.storage[index]
       } else {
+        // If the index is not empty, iterate through the bucket
         for (let i = 0; i < this.storage[index].length; i++) {
+          // If the key exists, delete the key-value pair
           if (this.storage[index][i][0] === key) delete this.storage[index][i]
         }
       }
     }
   }
 
+  // Check if a key exists in the hash table
   has(key) {
+    // Hash the key to find the index
     const index = this._hash(key, this.limit)
+    // Check if the bucket at the index exists
     if (this.storage[index]) {
+      // Iterate through the bucket's key-value pairs
       for (let i = 0; i < this.storage[index].length; i++) {
+        // Compare the current key with the target key and if the key is found, return true
         if (this.storage[index][i][0] === key) return true
       }
     }
+    // If the key is not found, return false
     return false
   }
 
+  // Print all keys/values in the table
   printTable() {
     for (let i = 0; i < this.storage.length; i++) {
       if (this.storage[i] !== undefined)
@@ -70,6 +98,7 @@ export default class HashTable {
     }
   }
 
+  // Clear all key/values
   clear() {
     this.storage = new Array(this.limit)
   }
--- a/Data-Structures/HashTable/HashTable.js
+++ b/Data-Structures/HashTable/HashTable.js
@@ -1,0 +1,76 @@
+export default class HashTable {
+  constructor(limit = 100) {
+    this.storage = new Array(limit)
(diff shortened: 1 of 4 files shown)


raklaptudirm (commented):
Needs way more comments to properly explain a concept like hash tables.

miladsade96:
@raklaptudirm OK

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
+export default class HashTable {
+  constructor(limit = 100) {
+    // Initialize the storage and limit variables
+    this.storage = new Array(limit)
+    this.limit = limit
+  }
+
+  // Hash function:
+  _hash(key, max) {
```
appgurueu, line 9: Use the [`#` syntax for private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties). The `max` parameter is also unnecessary; you can just get it as `this.storage.length` instead.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+    // If the index is empty, insert the key-value pair and create a bucket
+    if (this.storage[index] === undefined) this.storage[index] = [[key, value]]
+    else {
+      //  If the index is not empty, iterate through the bucket (collision handling)
+      let inserted = false
+      for (let i = 0; i < this.storage[index].length; i++) {
+        if (this.storage[index][i][0] === key) {
+          // If the key exists, update the value
+          this.storage[index][i][1] = value
+          inserted = true
```
appgurueu, line 31: Just early `return` here. Then you don't need the `inserted` variable, and you will be more efficient (by not doing any further loop iterations).

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+  // Insert a key-value pair into the hash table
+  set(key, value) {
+    // Hash the key
+    const index = this._hash(key, this.limit)
+    // If the index is empty, insert the key-value pair and create a bucket
+    if (this.storage[index] === undefined) this.storage[index] = [[key, value]]
+    else {
+      //  If the index is not empty, iterate through the bucket (collision handling)
+      let inserted = false
+      for (let i = 0; i < this.storage[index].length; i++) {
```
appgurueu, line 27: Use a `for`-`of` loop here: `for (const entry of this.storage[index])`.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
+export default class HashTable {
+  constructor(limit = 100) {
+    // Initialize the storage and limit variables
+    this.storage = new Array(limit)
```
appgurueu, line 4: It would be helpful to call this property `this.#buckets` (it should also be private).

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+    this.limit = limit
```
appgurueu, line 5: This field is redundant with `this.buckets.length`.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+    // Return the hash modulo the max
+    return hash % max
+  }
+
+  // Insert a key-value pair into the hash table
+  set(key, value) {
+    // Hash the key
+    const index = this._hash(key, this.limit)
+    // If the index is empty, insert the key-value pair and create a bucket
+    if (this.storage[index] === undefined) this.storage[index] = [[key, value]]
```
appgurueu, line 23: Use an object `{key, value}` here for better readability (s.t. instead of `[0]` and `[1]`, you can do `.key` and `.value` respectively). You could consider introducing a class `Entry` or `Pair` that stores a key-value pair for this.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+
+  // Get a value from the hash table
+  get(key) {
+    // Hash the key
+    const index = this._hash(key, this.limit)
+    // If the index is empty, return undefined
+    if (this.storage[index] === undefined) return undefined
+    else {
+      // If the index is not empty, iterate through the bucket
+      for (let i = 0; i < this.storage[index].length; i++) {
```
appgurueu, line 47: Use a `for`-`of` loop.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
+export default class HashTable {
+  constructor(limit = 100) {
```
appgurueu, line 2: 100 seems very arbitrary. I'd force the user to choose for a fixed-size hash table. (I'd also make it clear that this is a fixed-size hash table (with dynamically grown buckets) as opposed to a dynamic hash table. In particular, a fixed-size hash table can't guarantee expected O(1) operations.)

`limit` is also a slightly misleading name: This is not a hard limit, just a hash table "size". If you insert more elements, you will just get worse performance as your buckets fill up.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+  }
+
+  // Remove a key-value pair from the hash table
+  remove(key) {
+    // Hash the key
+    const index = this._hash(key, this.limit)
+    // Check if the bucket exists
+    if (this.storage[index]) {
```
appgurueu, line 59: I'd use an early return here: `const bucket = this.storage[index]; if (!bucket) return` (empty / `undefined` bucket). Then you don't need to indent the rest of the function.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+      // If the key matches the key at the index and there is only one item in the bucket, delete the bucket
```
appgurueu, line 60: (leaving an empty bucket behind wouldn't be an issue either; in fact i think the implementation may become slightly cleaner if you initialize all buckets to be empty, at a slight upfront performance cost at initialization)

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+      if (
+        this.storage[index].length === 1 &&
+        this.storage[index][0][0] === key
+      ) {
+        delete this.storage[index]
+      } else {
+        // If the index is not empty, iterate through the bucket
+        for (let i = 0; i < this.storage[index].length; i++) {
+          // If the key exists, delete the key-value pair
+          if (this.storage[index][i][0] === key) delete this.storage[index][i]
```
appgurueu, line 70: You're creating an array with holes here. I think this is rather dirty; it results in suboptimal time complexities (deleted items negatively affecting the time it takes to search a bucket).

It would be cleaner to just take an entry from the end, swap the entry to be deleted with that, then pop the entry from the end.

Data-Structures/HashTable/HashTable.js
```diff
@@ -0,0 +1,105 @@
...
+        // Compare the current key with the target key and if the key is found, return true
+        if (this.storage[index][i][0] === key) return true
+      }
+    }
+    // If the key is not found, return false
+    return false
+  }
+
+  // Print all keys/values in the table
+  printTable() {
```
appgurueu, line 93: I don't think "print" helpers are good design. A debugger, or even just `JSON.stringify`, work just as well for inspecting values.

A better abstraction would be an ["entries"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries)-like method to iterate over all the key-value-pairs.

appgurueu (commented):
Please remove unnecessary comments (prime example: `// Initialize the hash variable to 0` is plain redundant with the code).

The tests are very barebones. Ideally you should *roughly* try to reach decent coverage. I'd like to see some tests for collisions, for example.

A simple way to reach decent coverage would be to execute more or less random operations, comparing the result against JS's `Map` or a JS object.

PR #1641, opened 2024-03-08
feat: remove `twinPrime`

vil02:

### Describe your change:

This PR removes the function `twinPrime` as suggested in: https://github.com/TheAlgorithms/JavaScript/pull/1641#issuecomment-1986858870

~~This PR adds a test covering this line:~~ https://github.com/TheAlgorithms/JavaScript/blob/4a4ed57d4285fc61b5385d2c2525369ce2c4e6fb/Maths/TwinPrime.js#L23
~~And reorganizes existing tests.~~

### Checklist:


appgurueu (commented):
This improves coverage of the existing implementation, but I think the existing implementation has a bug: According to Wikipedia

> A twin prime is a prime number that is either 2 less or 2 more than another prime number

So the (a) twin prime of `7` should be `5`.

vil02:
> This improves coverage of the existing implementation, but I think the existing implementation has a bug: According to Wikipedia
> 
> > A twin prime is a prime number that is either 2 less or 2 more than another prime number
> 
> So the (a) twin prime of `7` should be `5`.

What should then be the `twinPrime(5)`?

I have to admit that I did not get this function is supposed to do. It would be more natural to have a function like `isTwinPrime`.

appgurueu:
> What should then be the `twinPrime(5)`?

I think either 3 or 7 would be fine.

We may also remove this function entirely, or replace it with an "is twin prime" variant as you suggest, IMO. It's just confusing (and trivial) as is.

vil02:
> > What should then be the `twinPrime(5)`?
> 
> I think either 3 or 7 would be fine.
> 
> We may also remove this function entirely, or replace it with an "is twin prime" variant as you suggest, IMO. It's just confusing (and trivial) as is.

I will remove it.

appgurueu (approved):

raklaptudirm (approved):

PR #1645, opened 2024-03-25
refactor: reduce code duplication in `FloodFill`

vil02:

### Describe your change:

This PR reduced the _code duplication_ and adds missing tests.

### Checklist:

### DIFF
--- a/Recursive/FloodFill.js
+++ b/Recursive/FloodFill.js
@@ -9,7 +9,7 @@
  * @see https://www.techiedelight.com/flood-fill-algorithm/
  */
 
-const neighbors = [
+const neighborOffsets = [
   [-1, -1],
   [-1, 0],
   [-1, 1],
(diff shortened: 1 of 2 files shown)


appgurueu (approved):
You could also consider making `checkLocation` return a boolean - it could return whether the location is valid.

Then you'd be duplicating the two `throw`s (but this is actually not that bad, because such side effects being "hidden" in a function may be a bit surprising).

You would then be able to deduplicate the `x >= 0 && x < rgbData.length && y >= 0 && y < rgbData[0].length` bounds check as well.

vil02:
@appgurueu thanks for 1be1853dd012718ec3b139ea3d1c48e29312fcdf.

appgurueu (approved):
Ah, I forgot to approve, didn't I...

raklaptudirm (approved):

PR #1668, opened 2024-06-09
test: add mixed number types for LocalMaximomPoint tests

taekim-dev:

### Describe your change:

- [ ] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Data-Structures/Array/test/LocalMaximomPoint.test.js
+++ b/Data-Structures/Array/test/LocalMaximomPoint.test.js
@@ -12,7 +12,6 @@ describe('LocalMaximumPoint tests', () => {
   })
 
   it('test boundary maximum points - should find first maximom point from the top', () => {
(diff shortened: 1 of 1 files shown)


appgurueu (commented):
I don't think these help coverage since the algorithm is independent of the specific numbers and just cares about the ordering.

If we're seeing this as contract-based testing, I also consider it a bit dubious that you expect a specific maximum when there are multiple valid return values (as is the case in your first test).

All in all, I'm afraid I wouldn't support this PR - I don't see the benefit it brings.

---

ps. you also seem to have unrelated changes in here, though that would be easy to fix.

taekim-dev:
@appgurueu Thank you for your feedback. Based on your insights, I have adjusted the test cases to align with the behavior of the existing algorithm.

I recognize the limitation of the original algorithm in not always returning the first local maximum. Your feedback on further improvements is welcomed.

PR #1676, opened 2024-07-08
Added Kann's algorithm

RaviSadam:
### Summary
- Kann's algorithm, a well known graph algorithm for finding the topological sorting order of graph

### File Added
- **KannsAlgoritm.js**: Kann's Algorithm implementation added in **Graphs** folder
- **KannsAlgorithm.test.js** : Test file for Kann's Algorithm added in Graphs/test folder


Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param n - number of vertices
+ * @returns {Array} - Empty array if cycle is detected or else result array;
+ *
+ */
+
+export function kannsAlgorithm(graph, n) {
+  if (n === null || n === undefined) throw Error('Invalid n was given')
+  const inorder = Array(n).fill(0)
+  const result = []
+  for (let entry of graph) {
```
appgurueu, line 19: A better name would be "neighbors" or something. Also should be `const`.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+    for (let edge of entry) {
```
appgurueu, line 20: Misleading name, this is a neighbor index (corresponding to an edge), not an edge. I suggest `const neighbor of neighbors`.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+      inorder[edge] += 1
+    }
+  }
+  const queue = []
+  console.log(inorder)
```
appgurueu, line 25: No side effects. Please remove this.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+export function kannsAlgorithm(graph, n) {
+  if (n === null || n === undefined) throw Error('Invalid n was given')
+  const inorder = Array(n).fill(0)
+  const result = []
+  for (let entry of graph) {
+    for (let edge of entry) {
+      inorder[edge] += 1
+    }
+  }
+  const queue = []
```
appgurueu, line 24: This need not be a queue. I recommend using it as a stack instead.

A better name would also inform readers that these are the nodes with no incoming edge.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param graph - Graph [[v1,v2],[v3,v4,v5]..]
+ * @param n - number of vertices
+ * @returns {Array} - Empty array if cycle is detected or else result array;
+ *
+ */
+
+export function kannsAlgorithm(graph, n) {
+  if (n === null || n === undefined) throw Error('Invalid n was given')
+  const inorder = Array(n).fill(0)
+  const result = []
```
appgurueu, line 18: (By the way, you could preallocate this array.)

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+  const queue = []
+  console.log(inorder)
+  for (let i = 0; i < n; i++) {
+    if (inorder[i] === 0) {
+      queue.push(i)
+    }
+  }
+  while (queue.length != 0) {
```
appgurueu, line 31: Our code style would prefer `!==` here. I think `>` is also a solid option, but `!=` is a bit odd.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+    const node = queue[0]
+    result.push(node)
+    queue.splice(0, 1)
```
appgurueu, line 34: Don't do this. This is potentially linear time. Just store the nodes without incoming edges in a stack instead, then you can pop from the end.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+    for (let nei of graph[node]) {
```
appgurueu, line 35: Again I'd recommend `const neighbor`. Could also call `node` something like `v`, then this could be called `w` if you wanted to be terse.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param graph - Graph [[v1,v2],[v3,v4,v5]..]
+ * @param n - number of vertices
+ * @returns {Array} - Empty array if cycle is detected or else result array;
+ *
+ */
+
+export function kannsAlgorithm(graph, n) {
+  if (n === null || n === undefined) throw Error('Invalid n was given')
+  const inorder = Array(n).fill(0)
```
appgurueu, line 17: Misleading name, a better name would be `indegree` or `indeg` for short.

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+      queue.push(i)
+    }
+  }
+  while (queue.length != 0) {
+    const node = queue[0]
+    result.push(node)
+    queue.splice(0, 1)
+    for (let nei of graph[node]) {
+      inorder[nei] -= 1
+      if (inorder[nei] == 0) {
```
appgurueu, line 37: Again `===`

Graphs/test/KannsAlgorithm.test.js
```diff
@@ -0,0 +1,13 @@
+import { kannsAlgorithm } from '../KannsAlgorithm'
+
```
appgurueu, line 2: This should all be in one `describe("Kahn's algorithm", ...)` block.

Graphs/test/KannsAlgorithm.test.js
```diff
@@ -0,0 +1,13 @@
...
+test('Test Case 1: Graph without cycle', () => {
```
appgurueu, line 3:
```suggestion
test('Graph without cycle', () => {
```

Graphs/test/KannsAlgorithm.test.js
```diff
@@ -0,0 +1,13 @@
...
+  const graph = [[], [], [3], [1], [0, 1], [0, 2]]
+
+  expect(kannsAlgorithm(graph, 6)).toEqual([4, 5, 0, 2, 3, 1])
+})
+
+test('Test Case 2: Graph with cycle', () => {
```
appgurueu, line 9:
```suggestion
test('Graph with cycle', () => {
```

appgurueu (changes requested):
It's "Kahn's algorithm", not "Kann's algorithm"

Graphs/KannsAlgorithm.js
```diff
@@ -0,0 +1,44 @@
...
+ * @author {RaviSadam}
+ * @name kannsAlgorithm
+ * @description -
+ * Kann's Algorithm implementation in JavaScript
+ * @summary
+ * Kann's Algorithm is used for topological sorting in directed acyclic graphs
+ *
+ * @param graph - Graph [[v1,v2],[v3,v4,v5]..]
+ * @param n - number of vertices
+ * @returns {Array} - Empty array if cycle is detected or else result array;
```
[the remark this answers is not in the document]
RaviSadam: Updated with return value with null in case of cyclic graph.

Graphs/KannsAlgorithm.js
RaviSadam: Modified variable names

Graphs/KannsAlgorithm.js
RaviSadam: Modified

Graphs/KannsAlgorithm.js
RaviSadam: Removed console.log

Graphs/KannsAlgorithm.js
RaviSadam: Preallocate could increase the time complexity. At the end we need to check array filled or not.

Graphs/KannsAlgorithm.js
RaviSadam: Modified

Graphs/KannsAlgorithm.js
RaviSadam: Uodated

Graphs/test/KannsAlgorithm.test.js
RaviSadam: Changes done

Graphs/test/KannsAlgorithm.test.js
RaviSadam: Done sir

Graphs/KannsAlgorithm.js
RaviSadam: Updated to inDegree

Graphs/KannsAlgorithm.js
RaviSadam: Updted with Queue defined in Data-Structures

Graphs/KannsAlgorithm.js
RaviSadam: I am using BFS algo. So have to use Queue data structure

Graphs/KannsAlgorithm.js
appgurueu: > Preallocate could increase the time complexity.

No, not really. It's still linear time in the size of the graph - O(n + m) where n is the number of nodes and m is the number of edges. (Allocation can also optimistically be assumed to be constant time.)

> At the end we need to check array filled or not.

Yes, and that can be done in constant time, if you keep an index to the last element. It'd look something like this:

```javascript
const topo_order = Array(n) // preallocate
const idx = 0
...
// to push:
topo_order[idx++] = node
...
if (idx < n) throw new Error("no topological order exists (graph is not acyclic)");
```

Or another variant to write this would be:

```javascript
const topo_order = Array(n) // preallocate
for (let i = 0; i < n; ++i) {
    if (stack.length === 0) throw new Error("no topological order exists (graph is not acyclic)");
    const node = stack.pop();
    topo_order[i] = node;
    ...
}
```

The benefit is that when a topological order exists (which should usually be the case; this is the "happy path"), you can save some reallocations of the array.

But this is just a suggestion, I'm fine with pushing to keep the code simple as well. Though perhaps the preallocation variant is a bit more readable, since it makes explicit that we want to have all n nodes be in this array.

Graphs/KahnsAlgorithm.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @author {RaviSadam}
+ * @name KahnsAlgorithm
+ * @description -
+ * Kahn's Algorithm implementation in JavaScript
+ * @summary
+ * Kahn's Algorithm is used for finding the topological sorting order of directed acyclic graph
+ *
+ * @param graph - Graph (adjacency list)
+ * @param n - Size of graph
```
appgurueu, line 12: Redundant. Just use `graph.length`.

Graphs/KahnsAlgorithm.js
```diff
@@ -0,0 +1,45 @@
...
+ * @returns {Array} - null if cycle is detected or else result array
+ *
+ */
+
+export function KahnsAlgorithm(graph, n) {
```
appgurueu, line 17: I'd prefer calling this `topoSort` or similar.

Graphs/KahnsAlgorithm.js
```diff
@@ -0,0 +1,45 @@
...
+  if (n === null || n === undefined) throw Error('Invalid n was given')
+  const inDegree = Array(n).fill(0)
+  const result = []
+  for (const neighbours of graph) {
+    for (const neighbour of neighbours) {
+      inDegree[neighbour] += 1
+    }
+  }
+  const queue = new Queue()
```
appgurueu, line 26: Why use a queue when you can get away with just using JS data structures, specifically, a stack?

Also, why call this "queue" rather than giving it a more descriptive name like "roots"?

Graphs/test/KahnsAlgorithm.test.js
```diff
@@ -0,0 +1,13 @@
+import { KahnsAlgorithm } from '../KahnsAlgorithm'
+
+test('Graph without cycle', () => {
```
appgurueu, line 3: As said, these two `test` blocks should be in a `describe("Kahn's algorithm", () => {...})` block.

appgurueu (changes requested):

RaviSadam:
Hai,
I am closing these pull requests as they involve substantial file modifications, resulting in an overly large number of commits.

PR #1714, opened 2024-10-05
feat: Add GaleShapley in a new folder Greedy

BKarthik7:

### Describe your change:

- [x] Add an algorithm?
- [x] Documentation change?

### Checklist:

### DIFF
--- a/Greedy/GaleShapley.js
+++ b/Greedy/GaleShapley.js
@@ -1,0 +1,78 @@
+/**
+ * The Gale-Shapley algorithm is used to find a stable matching between two sets
+ * of equal size (e.g., donors and recipients) where no two elements prefer each
+ * other over their current partners. It ensures a stable matching by proposing
+ * from one side to the other until all are matched.
+ *
+ * Complexity:
+ *    Worst-case performance O(n^2)
+ *    Best-case performance O(n^2)
+ *
+ * Reference:
+ *    https://en.wikipedia.org/wiki/Stable_marriage_problem
+ *    https://www.youtube.com/watch?v=Qcv1IqHWAzg (Numberphile)
+ *
+ */
+
+/**
+ *
+ * @param {number[][]} donorPref Preferences of donors, where each donor has an ordered list of recipients.
+ * @param {number[][]} recipientPref Preferences of recipients, where each recipient has an ordered list of donors.
+ * @returns {number[]} Array where the index is the donor, and the value at the index is the matched recipient.
+ *
+ * @example
+ * const donorPref = [[0, 1, 3, 2], [0, 2, 3, 1], [1, 0, 2, 3], [0, 3, 1, 2]];
+ * const recipientPref = [[3, 1, 2, 0], [3, 1, 0, 2], [0, 3, 1, 2], [1, 0, 3, 2]];
+ * stableMatching(donorPref, recipientPref); // Output: [1, 2, 3, 0]
+ */
+function stableMatching(donorPref, recipientPref) {
+  // Initialize the number of donors and create a list of unmatched donors
+  let n = donorPref.length
+  let unmatchedDonors = Array.from({ length: n }, (_, i) => i)
+
+  // Records of which recipient each donor is paired with and vice versa
+  let donorRecord = Array(n).fill(-1) // Donor to recipient
+  let recRecord = Array(n).fill(-1) // Recipient to donor
+
+  // Array to track how many recipients each donor has proposed to
+  let numDonations = Array(n).fill(0)
+
+  // While there are unmatched donors
+  while (unmatchedDonors.length > 0) {
+    // Take the first unmatched donor
+    let donor = unmatchedDonors[0]
+    let donorPreference = donorPref[donor]
+
+    // Find the next recipient this donor prefers
+    let recipient = donorPreference[numDonations[donor]]
+    numDonations[donor] += 1
+
+    // Get recipient's preference list and check the current match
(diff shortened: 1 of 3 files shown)


BKarthik7:
@appgurueu @raklaptudirm can you please review

BKarthik7:
@raklaptudirm can you please review

Greedy/test/GaleShapley.test.js
```diff
@@ -0,0 +1,36 @@
+import { expect, test } from 'vitest'
+import { stableMatching } from '../GaleShapley'
+
+test('Test Case 1', () => {
```
appgurueu, line 4: A test case structure like this isn't helpful. Please see https://github.com/TheAlgorithms/TypeScript/blob/master/CONTRIBUTING.md#writing-good-tests (it's for the TS repo, but translates to JS just as well).

TL;DR: Use a `describe` block for Gale-Shapley and `it.each` for individual test cases if you have no descriptive labels for them.

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {number[][]} recipientPref Preferences of recipients, where each recipient has an ordered list of donors.
+ * @returns {number[]} Array where the index is the donor, and the value at the index is the matched recipient.
+ *
+ * @example
+ * const donorPref = [[0, 1, 3, 2], [0, 2, 3, 1], [1, 0, 2, 3], [0, 3, 1, 2]];
+ * const recipientPref = [[3, 1, 2, 0], [3, 1, 0, 2], [0, 3, 1, 2], [1, 0, 3, 2]];
+ * stableMatching(donorPref, recipientPref); // Output: [1, 2, 3, 0]
+ */
+function stableMatching(donorPref, recipientPref) {
+  // Initialize the number of donors and create a list of unmatched donors
```
appgurueu, line 29: I'd use this array as a stack because that's efficient.

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+  let n = donorPref.length
+  let unmatchedDonors = Array.from({ length: n }, (_, i) => i)
+
+  // Records of which recipient each donor is paired with and vice versa
+  let donorRecord = Array(n).fill(-1) // Donor to recipient
```
appgurueu, line 34: I'd compute `donorRecord` at the end (or not at all even and instead just return `recRecord` and document that because the two are equivalent). That way you save yourself having to remember to update it during the algorithm, and the constant factor of the $O(n^2)$ many operations is better.

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+  let recRecord = Array(n).fill(-1) // Recipient to donor
+
+  // Array to track how many recipients each donor has proposed to
+  let numDonations = Array(n).fill(0)
+
+  // While there are unmatched donors
+  while (unmatchedDonors.length > 0) {
+    // Take the first unmatched donor
```
appgurueu, line 42: Don't. Take the last one instead; use the array as a stack (you could also use a queue, but why do that when a stack suffices). If you take the first one, you have to do a costly $O(n)$ unshift / splice later on to remove it (arguably you could also swap with the last one and pop, but that's just unnecessary complexity when you can just pop here), which wrecks your time complexity: You get $O(n^3)$ instead of $O(n^2)$.

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+    let donor = unmatchedDonors[0]
+    let donorPreference = donorPref[donor]
+
+    // Find the next recipient this donor prefers
+    let recipient = donorPreference[numDonations[donor]]
+    numDonations[donor] += 1
```
appgurueu, line 48:
```suggestion
    numDonations[donor]++
```

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+ * stableMatching(donorPref, recipientPref); // Output: [1, 2, 3, 0]
+ */
+function stableMatching(donorPref, recipientPref) {
+  // Initialize the number of donors and create a list of unmatched donors
+  let n = donorPref.length
+  let unmatchedDonors = Array.from({ length: n }, (_, i) => i)
+
+  // Records of which recipient each donor is paired with and vice versa
+  let donorRecord = Array(n).fill(-1) // Donor to recipient
+  let recRecord = Array(n).fill(-1) // Recipient to donor
```
appgurueu, line 35: I also think it'd be idiomatic to use `null` rather than `-1` to signal the absence of a value.

Greedy/GaleShapley.js
```diff
@@ -0,0 +1,78 @@
...
+    let recipient = donorPreference[numDonations[donor]]
+    numDonations[donor] += 1
+
+    // Get recipient's preference list and check the current match
+    let recPreference = recipientPref[recipient]
+    let prevDonor = recRecord[recipient]
+
+    // If recipient is already matched, check if they prefer the new donor
+    if (prevDonor !== -1) {
+      if (recPreference.indexOf(prevDonor) > recPreference.indexOf(donor)) {
```
appgurueu, line 56: This also wrecks time complexity by adding a linear factor. Instead you should precompute the inverse mappings initially as lookup tables (this is what's responsible for the $O(n^2)$ best case): For each receiver, compute the mapping from donor -> preference ahead of time.

appgurueu (changes requested):

PR #1728, opened 2024-10-10
feat: add new Josephus file in Recursive directory

Lonercode:

### Describe your change:

As part of my contribution to hacktoberfest 2024, I have added a file recursively implementing the Josephus algorithm and its associated test to the Recursive directory. The Josephus algorithm is a counting-out game that returns the last number after removing every kth number in a circle.  More on it can be found [here](https://en.wikipedia.org/wiki/Josephus_problem)

- [x] Add an algorithm?
- [ ] Documentation change?

### Checklist:

### DIFF
--- a/Recursive/Josephus.js
+++ b/Recursive/Josephus.js
@@ -1,0 +1,25 @@
+/**
+ * @function Josephus
+ * @description recursive implementation of the Josephus function.
+ * @param {Integer[]} collection - The integer array.
+ * @param {Integer} - The step size.
+ * @return {Integer} - The last integer in the list.
+ * @see [JosephusProblem](https://en.wikipedia.org/wiki/Josephus_problem)
+ * @example [1,2,3,4,5,6,7] with step 3 = 4
+ */
+
+const josephus = (collection, step) => {
+  // return null for invalid steps that are less than or equal to 0
+  if (step <= 0 || collection.length === 0) {
+    return null
+  }
+  if (collection.length === 1) {
+    return collection[0]
+  } else {
+    step = (step - 1) % collection.length
+    collection.splice(step, 1)
+    return josephus(collection, step + collection.length)
+  }
+}
+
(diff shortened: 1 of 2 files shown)


Recursive/Josephus.js
```diff
@@ -0,0 +1,25 @@
...
+const josephus = (collection, step) => {
+  // return null for invalid steps that are less than or equal to 0
+  if (step <= 0 || collection.length === 0) {
+    return null
+  }
+  if (collection.length === 1) {
+    return collection[0]
+  } else {
+    step = (step - 1) % collection.length
+    collection.splice(step, 1)
```
appgurueu, line 20: This is $O(n)$ so you get $O(n^2)$ runtime, which is very naive. Wikipedia shows formulae that achieve $O(n)$ or $O(k\log{n})$ runtime, both of which are substantially better. And even besides those, I can pretty easily come up with $O(n \log{n})$ algorithms based on decremental data structures which support this "splice" operation in logarithmic time (such as an initially balanced k-nary tree augmented with counts).

This algorithm also doesn't really benefit from recursion. It would be clearer as a simple `while` loop:

```javascript
while (collection.length > 1) { /* kill someone */ }
return collection[0]
```

Recursive/test/Josephus.test.js
```diff
@@ -0,0 +1,40 @@
+import { josephus } from '../Josephus'
+
+describe('Josephus', () => {
+  const collection = [1, 2, 3, 4, 5, 6, 7]
+  const collection1 = [1]
+  const collection2 = []
+
+  it('should return 4 for step size of 3', () => {
```
appgurueu, line 8: Repetitive tests of the form "should return X for Y" should use `it.each`

Lonercode:
Thanks for the review @appgurueu Alright. In that case would there be a need to modify it for the recursive directory.

